PR #1, opened 2021-05-12
Add a review action

btkostner:
This takes an action that when a new PR happens, will parse the new file, get all of the application information, then run the `bilelmoussaoui/flatpak-github-actions/flatpak-builder` action on it.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,13 +1,11 @@
 ---
 
-name: Review
+name: Publish
 
 on:
-  pull_request:
-    types:
-      - opened
-      - synchronize
-      - reopened
+  push:
+    branches:
+      - main
     paths:
       - applications/**/*.json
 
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,60 @@
...
+      source: ${{ steps.parse.outputs.source }}
+      commit: ${{ steps.parse.outputs.commit }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Parse
+        id: parse
+        uses: elementary/action-appcenter-review-parse@main
```
btkostner, line 29: This action takes all of the modified files in the PR, finds the one that matches the regex given (`applications/<rdnn>/<version>.json`) and then parses that file and merges it with the regex to get the `rdnn`, `version` (git tag / semver version), `source` (any valid public git repo), and git `commit`. It then runs the next step which is our standard flatpak building step.

cassidyjames (approved):
Let's go

PR #3, opened 2021-05-12
update README file

btkostner:
I'm going to need @cassidyjames to look over this and make sure everything looks good.


cassidyjames (approved):
I'll probably tweak some things in a follow-up but this is good to get rolling.

PR #31, opened 2021-07-21
GPLv3+ license added

comradekingu:
Needs approval from contributors.
If that includes me, I approve.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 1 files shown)


ryonakano:
Just out of curious, but what if a submitted app is licensed under another license, e.g. MIT? Or doesn't this LICENSE file just affect to the content of this repository and my curiosity is out of scope of this LICENSE file?

comradekingu:
@ryonakano If it already has a license, it can still be changed. If it is MIT you can change it into anything, because that is a weak license.

This time the whole repository is proprietary and everyone that committed to it owns their own contributions.
That means everyone has to agree to license if everyone is to use and it and for it to be compatible and so on and so forth.

PR #34, opened 2021-07-30
com.github.ryonakano.konbucase 3.2.0

danrabbit:
- Repository: [ryonakano/konbucase](https://github.com/ryonakano/konbucase)
 - Submitted By: @ryonakano


danrabbit:
@ryonakano it looks like the build here is failing because it's unable to resolve the submodule. It might be better to include the library as a module in your flatpak manifest instead of a submodule in your git project

ryonakano:
@danrabbit I created a new release [3.2.1](https://github.com/ryonakano/konbucase/releases/tag/3.2.1) with that fix, would you update the json and check if it now works?

danrabbit:
Works as expected, nice job!

PR #56, opened 2021-08-09
Check for JSON manifest

danrabbit:
Manifests can be either YML or JSON

I imagine there's probably a better way to DRY this?

publish also assumes YML, so need to copy whatever the final result here is to there

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -31,6 +31,27 @@ jobs:
           regex: |
             applications\/(?<rdnn>[a-zA-Z0-9\.\-]+)\.json
 
+      - name: Check for YML manifest
+        id: check_yml
+        uses: andstor/file-existence-action@v1
+        with:
+          files: "${{ needs.parse.outputs.rdnn }}.yml"
+
+      - name: Check for JSON manifest
+        if: steps.check_yml.outputs.files_exists == 'false'
+        id: check_json
+        uses: andstor/file-existence-action@v1
+        with:
+          files: "${{ needs.parse.outputs.rdnn }}.json"
+
+      - name: Create YML Manifest Path
+        if: steps.check_yml.outputs.files_exists == 'true'
+        run: echo "::set-output name=manifest::${{ needs.parse.outputs.rdnn }}.yml"
+
+      - name: Create JSON Manifest Path
+        if: steps.check_json.outputs.files_exists == 'true'
+        run: echo "::set-output name=manifest::${{ needs.parse.outputs.rdnn }}.json"
+
   publish:
     name: Publish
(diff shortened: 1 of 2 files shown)


btkostner (changes requested):
This works, but would mark the action as passed if neither a yaml or json file was found. It also needs to be applied to the publish action.

Ideally in the long run I will probably just make an action to find the bundle file in the repository. That way we don't have to have 2 build steps.

danrabbit:
@btkostner okay I think I figured out how to DRY the build step and I think this should probably fail if no manifest is found

.github/workflows/publish.yml
```diff
@@ -31,6 +31,27 @@ jobs:
           regex: |
             applications\/(?<rdnn>[a-zA-Z0-9\.\-]+)\.json
 
+      - name: Check for YML manifest
```
btkostner, line 34: We don't clone the remote repo here, so it's checking for the manifest file in the appcenter-reviews repo

danrabbit: derp my bad

btkostner: It's all good. I'll write a small task to get the file or fail

.github/workflows/review.yml
```diff
@@ -34,31 +33,45 @@ jobs:
...
       - name: Create JSON Manifest Path
         if: steps.check_json.outputs.files_exists == 'true'
-        run: echo "::set-output name=manifest::${{ needs.parse.outputs.rdnn }}.json"
+        run: echo '::set-output name=path::${{ needs.parse.outputs.rdnn }}.json'
 
   build:
     name: Build
     runs-on: ubuntu-latest
-    needs: parse
+    needs: [parse, manifest]
```
btkostner, line 74: This should only need `manifest` because `manifest` itself depends on `parse`.

.github/workflows/publish.yml
danrabbit: I took another crack at it. It doesn't seem to work in my fork, but I'm not sure why :dizzy_face:

btkostner (approved):
This _should_  work I think

btkostner:
Updated with new action I just made https://github.com/elementary/action-appcenter-review-find-manifest. I have not tested any of the JS I wrote so :grimacing:

danrabbit:
@btkostner copied this over to a fork to test, but it's failing: https://github.com/danrabbit/appcenter-reviews/runs/3305292656

btkostner:
@danrabbit I guess it was too positive that my JS code would work without actually testing it. `v1.0.3` is tested working

danrabbit:
Confirmed! Thanks for taking the time to fix that up :)

Issue #37, opened 2021-07-30
Action to add checklist to PR body


cassidyjames:
## Problem

We have to step through the same review steps on every PR which is hard to always remember. Plus, it's not as transparent to the public where a review is unless we comment as we review it.

## Proposal

Have an action to automatically throw a checklist into the PR body so we can check those off as it's reviewed.

## Prior Art

I've seen actions that comment on a PR with a checklist, but using the PR body is nice because GitHub considers it the overall progress for the PR in the header and on the PRs page.

cassidyjames:
Example list:

- [x] App opens
- [x] Does what it says
- [x] Matches description in AppData
- [x] Matches screenshot
- [x] AppData custom colors are contrasty
- [x] Name is unique and non-confusing

There's probably more to check, but this is a start.

PR #60, opened 2021-08-12
Add checklist setup for PRs

btkostner:
Should fix https://github.com/elementary/appcenter-reviews/issues/37

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: Checklist
-        uses: elementary/action-appcenter-review-checklist
+        uses: elementary/action-appcenter-review-checklist@v1.0.0
         with:
           body: |
             ## Review Checklist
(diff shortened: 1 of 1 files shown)


.github/workflows/review.yml
```diff
@@ -12,6 +12,24 @@ on:
...
+        with:
+          body: |
+            ## Review Checklist
+
+            - [ ] App opens
+            - [ ] Does what it says
+            - [ ] Matches description in AppData
+            - [ ] Matches screenshot
+            - [ ] AppData custom colors are contrasty
+            - [ ] Name is unique and non-confusing
```
btkostner, line 31: The markdown is configurable, but this is the list that Cass gave so I think it's a pretty good starting point. Easily editable down the road.

cassidyjames (approved):
Let's do it. Is there a good way to test this, or just merge in and then watch new PRs? :)

.github/workflows/review.yml
cassidyjames: Yeah I was thinking the same. I'm probably missing things but we can iterate as we review things and learn what we need to explicitly check.

btkostner:
Merge and watch :P

I did add tests to this package, so I'm pretty confident it works

PR #73, opened 2021-08-13
com.github.krisives.videodownloader 0.8.6

danrabbit:
- Repository: [krisives/videodownloader](https://github.com/krisives/videodownloader)
 - Submitted By: @krisives

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Matches description in AppData
- [ ] Matches screenshot
- [ ] AppData custom colors are contrasty
- [ ] Uses elementary Flatpak runtime

### DIFF
--- a/applications/com.github.krisives.videodownloader.json
+++ b/applications/com.github.krisives.videodownloader.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/krisives/videodownloader.git",
+  "commit": "eed87a97e4188789227787377a4f0bb5976600c7",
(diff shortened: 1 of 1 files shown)


danrabbit:
Issues are disabled on the repository, so I'll have to leave feedback here

@krisives, forks aren't explicitly disallowed in AppCenter, but you will need to use your own trademarks. You can't redistribute someone else's app with the same name, icon, etc and your app should have some significant difference as a fork. Forking isn't a way to get around apps needing to be maintained by their author

PR #82, opened 2021-08-18
com.github.tenderowl.frog 0.1.1

btkostner:
Repository: [TenderOwl/Frog](https://github.com/TenderOwl/Frog)
Submitted By: @amka

### DIFF
--- a/applications/com.github.tenderowl.frog.json
+++ b/applications/com.github.tenderowl.frog.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/TenderOwl/Frog.git",
+  "commit": "f7e70b39e3df9e733839744c61253014a78d78ba",
+  "version": "0.1.1"
+}


amka:
Huh, seems like the problem the manifest is placed under the "flatpak" directory, not the root of the project.

cassidyjames:
@btkostner it looks like the `flatpak-builder` action accepts a `manifest-path` input https://github.com/bilelmoussaoui/flatpak-github-actions/blob/69e09d1378a7d2e492c07f7f45a71293383fc9c3/flatpak-builder/action.yml#L8

Should we do something here in the JSON, and then if that's missing, fall back to the regex?

btkostner:
@cassidyjames If we are going to include the manifest path in the JSON, I'd just drop the regex

cassidyjames:
@amka filed at https://github.com/elementary/action-appcenter-review-find-manifest/issues/2

btkostner:
https://github.com/elementary/appcenter-reviews/pull/86 should fix the issue

cassidyjames (changes requested):
https://github.com/TenderOwl/Frog/issues/5

amka:
Fixed in master and 0.1.2.

danrabbit:
The icon is not showing for the other app in AppCenter or on appCenter.elementary.io. If it's installed in the same way in this app, we should probably put this on hold until we figure out if it's an issue with the app or with AppCenter

cassidyjames:
@danrabbit ah that might be this which was missed https://github.com/TenderOwl/Turtle/pull/4; similarly, I've filed https://github.com/TenderOwl/Frog/pull/6

amka:
@cassidyjames  TenderOwl/Turtle#4, TenderOwl/Frog#6 merged.

amka:
@cassidyjames looks like fixed app ID didn't help :(

danrabbit:
@amka My guess is that it's this line here: https://github.com/TenderOwl/Frog/blob/1df23aab26f5af3e07de0c3ec603d49a737612cd/data/com.github.tenderowl.frog.appdata.xml.in#L31

It doesn't look like you're actually setting that to anything in your build system

amka:
[0.1.5](TenderOwl/Frog/releases/tag/0.1.5) pushed with new release tag and keywords by your suggestion. Thank you again @danrabbit  :)
Hash: `1fb07b838c4f41964c45492673dd60a9a24fcb37`

danrabbit (approved):
As long as CI passes, I think this is good to ship. Nice work!

PR #98, opened 2021-08-24
com.github.ztefn.haguichi 1.145.2

ztefn:
Repository: [ztefn/haguichi](https://github.com/ztefn/haguichi)
Submitted By: @ztefn
Release Tag: [1.145.2](https://github.com/ztefn/haguichi/releases/tag/1.145.2)

## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Matches description in AppData
- [x] Matches screenshot
- [x] https://github.com/ztefn/haguichi/issues/46
- [x] Uses elementary Flatpak runtime
- [x] https://github.com/ztefn/haguichi/issues/47

### DIFF
--- a/applications/com.github.ztefn.haguichi.json
+++ b/applications/com.github.ztefn.haguichi.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ztefn/haguichi.git",
-  "commit": "ec57125c1f6e634d80f92030a172631558232938",
-  "version": "1.145.1"
+  "commit": "07e8ae3e1fa059b3cbc42067e14b825da99c58a4",
+  "version": "1.145.2"
 }
--- a/applications/com.github.ztefn.haguichi.json
+++ b/applications/com.github.ztefn.haguichi.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ztefn/haguichi.git",
-  "commit": "b80ef34a13d3ad2a7e4a27f1a063ce1328d3d997",
-  "version": "1.145.0"
+  "commit": "ec57125c1f6e634d80f92030a172631558232938",
+  "version": "1.145.1"
 }
--- a/applications/com.github.ztefn.haguichi.json
+++ b/applications/com.github.ztefn.haguichi.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/ztefn/haguichi.git",
+  "commit": "b80ef34a13d3ad2a7e4a27f1a063ce1328d3d997",
+  "version": "1.145.0"
+}


danrabbit:
@ztefn Before I go and file a bunch of issues on your repository, it looks like your app doesn't meet many of our submission guidelines detailed here: https://docs.elementary.io/develop/appcenter/publishing-requirements

Your app doesn't use the elementary flatpak runtime, your screenshots don't show the default stylesheet, etc.

Is it your intention to publish an app for elementary OS or is this app more targeted at GNOME and maybe better suited for publishing on Flathub?

cassidyjames:
@danrabbit make sure you're looking at the right tag; it looks like the elementary version is maintained in a separate branch that is tagged for the releases: https://github.com/ztefn/haguichi/tree/1.145.0

cassidyjames:
Hm, I'm greeted by this when first running. With Flatpak, is it possible to bundle the Hamachi client or at least automate the download instead of throwing the user out to a web page?

Screenshot from 2021-08-24 21-02-19@2x

I guess the issue is that Hamachi is required to be running on the host system, and can't just run in the container?

ztefn:
> I guess the issue is that Hamachi is required to be running on the host system, and can't just run in the container?

Yes, I'm pretty sure Hamachi would be pretty useless running inside a container (if it even technically could, which I'm doubtful) as it creates a network interface that you want have available when running other applications, such as Terminal (ping, ssh) or games (playing LAN mode over the internet).

> With Flatpak, is it possible to bundle the Hamachi client or at least automate the download instead of throwing the user out to a web page?

Automating the download of the .deb file and then installing it using PackageKit like [Eddy](https://github.com/donadigo/eddy) could be an interesting take, depending on how much effort it would be. After all Haguichi's primary role is providing a graphical frontend for Hamachi, not installing packages (insert Unix philosophy).

ztefn:
> it looks like the elementary version is maintained in a separate branch that is tagged for the releases

To be precise, all metadata patches needed for elementary OS (appdata, screenshots, flatpak manifest) are stacked on top of master branch in [elementary](https://github.com/ztefn/haguichi/tree/elementary) branch by using git rebase. Then tags for elementary OS are created from this branch.

Any actual code differences for elementary OS (like using Granite for dialogs) are maintained in master, but put behind the `-Dfor-elementary` build option.

hanaral:
I'd like to chime in to ask something that may help - how does eddy install debs as a flatpak? it may be extra effort, but copying it to install the deb may be a solution applicable to the flathub version as well.

danrabbit:
@ztefn thanks for answering questions and apologies for the confusion regarding branches.

I think this generally looks good to me. We probably don't want apps to directly access the login interface in the future just in case, but the use case here looks legit and I filed an issue report about using Gtk.Application's native methods instead.

@cassidyjames I'm wondering if we have precedent regarding the notifications interface? I filed an issue because it looks like Haguichi is using LibNotify and not sending a desktop-entry hint which breaks notification settings and won't let us group in the indicator. I'm inclined to say that should probably be addressed before publishing so users can not have to worry about if apps in AppCenter work with notifications or not.

Maybe not necessary for this app submission specifically, but I wonder if we should disallow using the notifications dbus interface manually altogether and require that apps use the notifications portal either via GLib.Application or LibPortal?

danrabbit (changes requested):
Alright, just copying in here since Cass responded in Slack:

I am going to request changes on this submission for the notifications bit. The notifications just need to set the `desktop-entry` hint so that it works with settings and the notification indicator.

danrabbit (approved):
Excellent, thank you so much for making that adjustment! As long as CI passes, I'm happy to publish this update in OS 6. Nice work! :tada:

PR #118, opened 2021-09-07
com.github.tenderowl.turtle 0.2.0

danrabbit:
https://github.com/TenderOwl/Turtle/compare/0.1.5...0.2.0

## Review Checklist

- [ ] App opens
- [ ] Does what it says

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/com.github.tenderowl.turtle.json
+++ b/applications/com.github.tenderowl.turtle.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/TenderOwl/Turtle.git",
-  "commit": "c2642da47a7cb40c6b61f3d8021a102be2dd0f9d",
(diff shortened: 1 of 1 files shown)


danrabbit:
@tenderowl can you explain what this new x-run-args bit is for with the reference to Godot in your manifest? It seems to be the only reference to Godot that I see

amka:
@danrabbit Oh, sure. X-args is used by the Builder to run app with arguments for testing purposes. And the Builder save it to the manifest. Nothing special.

danrabbit:
Thanks for supporting the dark style. Looks good!

PR #122, opened 2021-09-08
com.github.syfds.soundy 0.3.4

danrabbit:
Repository: [syfds/soundy](https://github.com/syfds/soundy)
Submitted By: @syfds

## Review Checklist

- [x] App opens
- [ ] Does what it says

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.syfds.soundy.json
+++ b/applications/com.github.syfds.soundy.json
@@ -1,5 +1,5 @@
 {
-  "source": "https://github.com/calo001/fondo.git",
+  "source": "https://github.com/syfds/soundy.git",
   "commit": "fa0d5b732acad2c448e9b7d25122aec121c98e05",
   "version": "0.3.3"
 }
--- a/applications/com.github.syfds.soundy.json
+++ b/applications/com.github.syfds.soundy.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/calo001/fondo.git",
+  "commit": "fa0d5b732acad2c448e9b7d25122aec121c98e05",
+  "version": "0.3.3"
+}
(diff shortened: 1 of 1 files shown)


danrabbit:
@syfds Hey thanks for submitting your app! I have a couple of questions :)

In AppData, you have listed `money-purchasing` as `intense`. I don't see anything in your screenshots that would indicate that your app includes purchases. Was this intentional? For reference, you can generate OARS data from https://hughsie.github.io/oars/generate.html

You also have pretty aggressive permissions for filesystem and dbus access:

```
  - '--socket=system-bus'
  - '--socket=session-bus'
  - '--filesystem=host'
```

Is it possible to scope these down a bit? For example, I don't see why this app needs filesystem permission at all as it doesn't seem to export or import any files.

syfds:
hi @danrabbit, thanks for your review! 

`money-purchasing` : the attributes were copy-pasted from an other project, so I adjusted each attribute to `none`. https://github.com/syfds/soundy/commit/cf4cbb483c621542fb4af73076ada0c17c6d78a1

Permissions: I tested a lot to be able to connect to .local dns names thats why a lot of not really needed permissions are there. A need it to be able to use the auto discovery feature. Sadly it is not possible for now with flatpak packaging (https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1013). I left only required permissions: https://github.com/syfds/soundy/commit/3746f8b306ed720d83844e8dd7f061cba1ae2a54

Should I create a new tag and publish it over https://beta.developer.elementary.io/submissions?

danrabbit:
@syfds I think you want to keep `- '--socket=wayland'` and `- '--share=ipc'` probably, but otherwise thanks for checking those permissions for me!

For OARS data, you don't have to explicitly declare `none` if you don't want to. You can just omit that tag and we assume that means `none`

You don't have to do anything with the dashboard website, just make a release tag when you're ready and I'll update the PR here :)

syfds:
hi @danrabbit , I created a new release 0.3.4: https://github.com/syfds/soundy/tree/0.3.4

danrabbit:
Looks good, thanks for making those changes!

PR #129, opened 2021-09-14
com.github.alexkdeveloper.raddiola 1.0.0

alexkdeveloper:
Repository: [alexkdeveloper/raddiola](https://github.com/alexkdeveloper/raddiola)
Submitted By: @alexkdeveloper

## Review Checklist

- [x] App opens
- [x] Does what it says

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime
- [x] https://github.com/alexkdeveloper/raddiola/issues/4


danrabbit:
Hey thanks for your submission! @alexkdeveloper can you explain why your app is requesting `--filesystem=home` permissions? I don't see where your app needs to read or write files to a user's home folder

danrabbit:
Ah, I think I see it. It looks like you're trying to write to a toplevel hidden folder in the user's home folder, but instead you probably want to use [`Environment.get_user_data_dir ()`](https://valadoc.org/glib-2.0/GLib.Environment.get_user_data_dir.html) which should return a sandboxed location your app has access to

alexkdeveloper:
Hi! Updated the json file. Added workflows. Packages are being collected!

alexkdeveloper:
Hi! I'm still an inexperienced github user, so something may not work out right away) I have updated the data.

alexkdeveloper:
I changed the permissions in json. Instead of filesystem=home to persist=. stations_for_radio_app.

alexkdeveloper:
Stations are not created at startup. New ones are created through the program interface.

alexkdeveloper:
My application works with this permission: 
--filesystem=home/. stations_for_radio_app 
Can I use it?

danrabbit (changes requested):
@alexkdeveloper no problem :)

It looks like your app is still requesting full home permissions. You'll need to either use gsettings or switch to ` Environment.get_user_data_dir ()` as pointed out above

alexkdeveloper:
Hi! I applied Environment.get_user_data_dir (). Removed the filesystem=home permission.

danrabbit (approved):
Thanks for making those changes. Looks good!

PR #134, opened 2021-09-18
com.github.ztefn.haguichi 1.145.3

ztefn:
Repository: [ztefn/haguichi](https://github.com/ztefn/haguichi)
Submitted By: @ztefn
Release Tag: [1.145.3](https://github.com/ztefn/haguichi/releases/tag/1.145.3)
Changed Code: [1.145.2...1.145.3](https://github.com/ztefn/haguichi/compare/1.145.2...1.145.3#diff-22391d6974c007b8c4f83f47f13c077c1c6a23c96941e780493be79a536bd42c)

This is a minor update that pins all GTK settings (stylesheet, icon theme, etc) so that if someone installs the flatpak from AppCenter on another distro everything looks as intended.

### DIFF
--- a/applications/com.github.ztefn.haguichi.json
+++ b/applications/com.github.ztefn.haguichi.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ztefn/haguichi.git",
-  "commit": "07e8ae3e1fa059b3cbc42067e14b825da99c58a4",
-  "version": "1.145.2"
+  "commit": "b55912bf6a759316055813863f49cf697f646a60",
+  "version": "1.145.3"
 }


danrabbit (approved):
In a future version you may want to consider doing something like checking if the theme prefix is `io.elementary.stylesheet` before hardcoding the blueberry stylesheet. This way users could still select accent colors :)

ztefn:
> In a future version you may want to consider doing something like checking if the theme prefix is `io.elementary.stylesheet` before hardcoding the blueberry stylesheet. This way users could still select accent colors :)

Haguichi [sets its own red accent color](https://github.com/ztefn/haguichi/blob/master/data/stylesheets/elementary-dark.css#L8).

PR #139, opened 2021-09-21
com.github.alexkdeveloper.dfc 1.0.0

alexkdeveloper:
## Review Checklist

- [x] App opens
- [x] Does what it says

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -37,7 +37,7 @@ jobs:
     needs: parse
 
     container:
-      image: ghcr.io/elementary/flatpak-platform/runtime:daily
+      image: ghcr.io/elementary/flatpak-platform/runtime:6
       options: --privileged
 
     steps:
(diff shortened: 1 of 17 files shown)


danrabbit (changes requested):
Hey @alexkdeveloper thanks for your submission and for being patient! Apologies I haven't gotten to your review sooner. I just have a couple of questions:

I see that you have network access listed in your manifest but I'm not sure what part of your app needs access to the network. Is this intentional?

I'm also wondering why your app needs read access to all of `home`. Could this be scoped down a bit?

Just a recommendation, but I would advise removing "Gnu/Linux" from your app description. This isn't a unique selling point for your app since all apps in AppCenter only run on linux-based operating systems. You might instead want to explain what desktop files are and/or why someone would want to create them

alexkdeveloper:
The permissions specify read-only access for the home folder. Otherwise, file selection does not work without it.

alexkdeveloper:
Removed network access from permissions. Changed the description.

PR #140, opened 2021-09-21
com.github.jmoerman.goforit

JMoerman:
This is my submission for _GoForIt!_ (formerly _Go For It!_)

I have submitted this 6 days prior via https://beta.developer.elementary.io/submissions, but that didn't appear to do much. (And it currently is offline.) If that submission is being processed please close this PR.

Things of note: I created the https://github.com/JMoerman/GoForIt repository for the AppCenter packaging files. The code is still hosted on https://github.com/JMoerman/Go-For-It. The reasons for this are:

- _GoForIt!_ is gradually splitting into a Gnome(/more desktop agnostic) and a elementary OS version. The differences are currently relatively small but will grow with time. I prefer not to have a flatpak manifest with a generic App-ID name in the code branch. (I prefer to keep packaging files out of the code branch altogether.)
- _Go For It!_ has been renamed to _GoForIt!_ as appstream (for example `appstreamcli`) will discard `go`, `for` `it` from the search string (it will keep `it!`) as these are all common words. Searching for `Go For It` will thus not give any useful results in AppCenter. _GoForIt!_ can at least be found by searching for  `GoForIt`, so the new situation should be at least a little better. As the name has changed, users will not upgrade from elementary OS 5 to 6 without reinstalling 3rd party apps, and the old App-ID contained hyphens I also want to rename the App-ID to `com.github.jmoerman.goforit`.

If you have suggestions on how to handle these issues, I would be happy to hear them.

### DIFF
--- a/applications/com.github.jmoerman.go-for-it.json
+++ b/applications/com.github.jmoerman.go-for-it.json
@@ -1,5 +1,0 @@
-{
-  "source": "https://github.com/JMoerman/Go-For-It.git",
-  "commit": "f626f185bc58f08f9d73afb2865e7acc6f16f825",
-  "version": "1.9.3"
-}
--- a/applications/com.github.jmoerman.go-for-it.json
+++ b/applications/com.github.jmoerman.go-for-it.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/JMoerman/Go-For-It.git",
+  "commit": "f626f185bc58f08f9d73afb2865e7acc6f16f825",
+  "version": "1.9.3"
+}
--- a/applications/com.github.jmoerman.goforit.json
+++ b/applications/com.github.jmoerman.goforit.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/JMoerman/GoForIt.git",
+  "commit": "9300ee4a2478f1dc7262de62abf80eb546033d8b",
+  "version": "1.9.4"
+}


danrabbit (changes requested):
Hey @JMoerman thanks for your patience and apologies for not being able to review your app sooner!

It looks like your app is failing to build due to a reference to a manifest file that doesn't exist in your repository. Specifically `libcanberra.json`

If you're using GLib.Notification (which as far as I can tell you are), you don't need to add a sandbox hole for freedesktop notifications on dbus, since it will automatically use the notifications portal :)

Does GoForIt need access to the users' entire `home`? I don't see anything immediately obvious where GoForIt would need any file access outside of its own sandbox. Can you elaborate on this?

JMoerman:
> Hey @JMoerman thanks for your patience and apologies for not being able to review your app sooner!

No problem!

> It looks like your app is failing to build due to a reference to a manifest file that doesn't exist in your repository. Specifically `libcanberra.json`

It is in the `shared-submodules` (from flathub) git submodule. I assumed that git submodules would be supported, which apparently isn't correct. This is easily fixed.

> If you're using GLib.Notification (which as far as I can tell you are), you don't need to add a sandbox hole for freedesktop notifications on dbus, since it will automatically use the notifications portal :)

I didn't realize that it wasn't required any more. I was using libnotify before and back then, I believe it was required.

> Does GoForIt need access to the users' entire `home`? I don't see anything immediately obvious where GoForIt would need any file access outside of its own sandbox. Can you elaborate on this?

Right, `xdg-desktop-portal` seems to work well enough by now and I do not need to support prior installations that used this permission, so I can drop this permission now. I do need to make some minor adjustments to make everything work without access to the home directory. I just noticed a pretty severe bug (not dangerous, but it will confuse/annoy users) so I will also need to tackle that issue too.

I will let you know when I have fixed these issues and updated this PR.

danrabbit:
Great, thank you! Looking forward to getting this app back in the store :)

JMoerman:
@danrabbit `libcanberra.json` is now stored in the repo itself and I have removed the `talk-name=org.freedesktop.Notifications` and `filesystem=home` permissions. Aside from that, the update to 1.9.5 should fix some issues and it should look slightly better.

PR #142, opened 2021-09-23
com.github.alexkdeveloper.somafm 1.0.0

alexkdeveloper:


danrabbit:
Hey @alexkdeveloper thanks for being patient and apologies for taking a while to review! Can you clarify the difference between this app and https://github.com/alexkdeveloper/raddiola ? They seem to be quite similar

alexkdeveloper:
Hi! This app is created for SomaFM lovers. It contains all 38 stations. Created on the basis of Raddiola.

alexkdeveloper:
This application also differs from Raddiola in that a text label with the name of the current station has been added. Also, a button to go to the site has been added to the title of the window somafm.com . The interface on the station editing page has been slightly changed. I think this app should appeal to all SomaFM lovers.

cassidyjames (changes requested):
@alexkdeveloper thanks for your submission to AppCenter. However, upon review I still don't understand why this is two apps. I am inclined to encourage you to find a way to integrate these stations into the existing app, as right now it could be confusing for users as to what the difference is and why there are two nearly identical apps from the same developer.

It also appears that [SomaFM](https://somafm.com/) is a brand name that you do not own; according to the [AppCenter Submission Guidelines](https://docs.elementary.io/develop/appcenter/publishing-requirements):

>**You may not publish your app using trademarks reserved by others**. …To protect both users and developers, **your app's name must not be the same as** or confusingly similar to an existing app in AppCenter or **another brand**, as determined by app reviewers.

We will also be updating the AppCenter submission guidelines for more clarity; while this app submission seems genuine, it is also important to prevent someone from spamming AppCenter with duplicate apps to try and flood a specific category or type of app.

PR #143, opened 2021-09-23
com.github.tenderowl.norka 0.8.0

amka:
- Submitted By: @amka 
- Repository: [TenderOwl/Norka](/TenderOwl/Norka.git)


danrabbit (changes requested):
Hey @amka thanks for your patience and apologies about taking a while to review!

I noticed that your app manifest has permissions for all of `home`. Is it possible to scale that down to just `documents`?

Could you also please remove the colored background from your screenshot?

Other than that I think we're good to go :)

PR #153, opened 2021-10-04
Build for ARM

danrabbit:


btkostner:
This assumes every project will be able to be built on that architecture. That might not be the case with dependencies and what not. I would think we should add an array to each project file (with a sane default) that makes the arch configurable

davidmhewitt:
Because of the `fail-fast: false`, this should still actually build and publish for x86, even if the ARM build fails.

I think a failing ARM build is also a useful canary for apps that might be trying to do weird stuff. If it doesn't build for ARM, why? These are sandboxed user-facing GUI apps that shouldn't be doing any stuff that's low level enough to care about what CPU architecture or endianness it's running on.

I guess there's the argument that an app could be written in a language that's trying to pull in binary dependencies, and those binary dependencies don't have an aarch64 version. But again, that's a sign to look closer and wonder why. Or equally, there could be an argument that an app developer doesn't want to publish their apps to ARM devices for some reason. But I don't think there would be many valid reasons for that.

I think having it configurable in the long term is probably a reasonable goal, but it looks like this would probably need some work on the `elementary/action-appcenter-review-parse` action to add that config key, so its whether there's anyone around that has time to do that, or whether we should just use this in the short term 🤷

danrabbit:
Yeah, I agree with David here. I don't think there really should be a reason that AppCenter apps wouldn't build for ARM. If they fail an ARM build, that's a problem worth investigating. If we're going to be trying to ship ARM devices like Pinebook, we should be aiming to guarantee the same user experience, including AppCenter

PR #183, opened 2021-10-30
com.github.dr_styki.screenrec 3.0.0

dr-Styki:
ScreenRec is Flatpak ready ! 😁️

## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] https://github.com/dr-Styki/ScreenRec/issues/47
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime
- [x] https://github.com/dr-Styki/ScreenRec/issues/48
- [x] https://github.com/dr-Styki/ScreenRec/issues/49


dr-Styki:
The screenshots and their URLs have been updated. 🙂️

danrabbit (changes requested):
Left a couple issue reports regarding the flatpak manifest, otherwise looks good!

dr-Styki:
I remove the unnecessary Flatpak sandbox holes. Unfortunately I can't remove the plugin, without it MP4 recording doesn't work

danrabbit:
@dr-Styki I don't think we can ship the "ugly" set of plugins for legal reasons. The description of the package from its developers state that it may contain codecs which are patent incumbered. Redistributing these in our app store could open elementary up to litigation.

danrabbit (changes requested):
The "ugly" plugins need to be removed so we can comply with US patent law

danrabbit (approved):
A couple of comments, but it doesn't block this release:

A new version of our Flatpak runtime 6.1 is available and we recommend you upgrade to that from 6 in your Flatpak manifest

`Recorder` seems to also be an appropriate category in your Desktop Entry as per: https://specifications.freedesktop.org/menu-spec/latest/apas02.html

Thank you! Excited to ship this!

PR #198, opened 2021-11-13
Create com.github.leggettc18.leopod

leggettc18:
I would like to publish my app for Elementary OS 6. I think I managed to accidentally submit it for review for Elementary OS 5.1, but I don't think it will build for that so that can be ignored. The documentation led me there before I found this.

## Review Checklist

- [x] App opens
 - [x] Does what it says
 - [x] Categories match
 - [x] https://github.com/leggettc18/leopod/pull/4
            
### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] https://github.com/leggettc18/leopod/issues/3
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime


leggettc18:
I just noticed I forgot to add the file with the .json extension. I was wondering why some parts of my pull request looked different from others! Sorry about that.

danrabbit (changes requested):
Hey @leggettc18 thanks for your submission! Sorry about the confusion with the dashboards. Things are a bit of a mess right now and we really need to get that sorted.

I noticed that you have a stripe key listed here. Is this the stripe key you received after connecting your account with the old dashboard through Stripe connect? If this is the public key for stripe account and not specifically for this app through stripe connect, payments will fail for your app.

Can you also explain why your app needs to poke holes in the sandbox for GVFS?

I made a pull request to add the `Feed` category to your desktop file. This isn't required to publish, but could help your app be discovered when we add more detailed categories to AppCenter.

This also isn't a required change, but just as an advice I would recommend using the native Gtk widgets for gstreamer instead of using Clutter.

leggettc18:
OK I'll look into the things you mentioned.

I needed GVFS sandbox exceptions in order for downloading the audio files and coverart to work using GLib.File.copy_async, even when downloading the files to the app's sandboxed folders for some reason. Is there something else I should be using to download these files? I searched for it a bit when I first flatpaked the app and discovered it wouldn't download anything but that search just led me to the GVFS exception that I did.

leggettc18:
I've been looking further into the GVFS situation to try and find an alternative with no success so far, Maybe I'm just searching for the wrong thing?

leggettc18:
Incidentally, I found I could make it work with `--socket=session-bus` but if flatpak's documentation is to be believed that seems to be even worse.

danrabbit:
Thanks for researching this! I'll make sure to set aside some time at the start of my day to see if I can find a solution for you! if not, we can live with the sandbox hole until we figure it out 😉

leggettc18:
I've made some progress towards figuring this out. Part of my code for downloading the episodes did a `remote_episode.query_exists ()` to check if the remote file actually existed before beginning the download. I removed that check and next visible part of the code (the part where the download details box appears) does execute without GVFS.  However, the download itself still appears to not actually start, with no other errors. So there is at least progress!

leggettc18:
I opened an issue on my end of the repo https://github.com/leggettc18/leopod/issues/5 to keep the results of my experimenting over on my repo's side rather than having it litter this PR.

danrabbit:
Hey @leggettc18 it sounds like we need to do some more research to resolve that particular issue, and I don't want to hold you up on it. So if you want to make a new release, let's go ahead and get this published and we can revisit the other issue at another time :)

danrabbit:
@leggettc18 Hm regarding the stripe key, I'm not able to confirm that here. It's possible that this is a valid key. Normally what would happen is you would get the key from AppCenter Dashboard, not from your stripe dashboard. I'll reach out to someone about this to confirm. In the meantime, we can try publishing this the way it is and then we can always follow up if this doesn't work

leggettc18:
Works for me. I'm good to go if you are.

danrabbit:
@leggettc18 as far as I can see on my end, it looks like the payment completed. Can you confirm that for me? :)

leggettc18:
Yep, I got the email from Stripe that I received it. So I guess that works! I wasn't seeing any key on the old dashboard and obviously I'm not able to see the new dashboard so I'm glad that key worked out.

PR #201, opened 2021-11-16
com.github.manexim.typewriter 0.6.1

meisenzahl:
Repository: [manexim/typewriter](https://github.com/manexim/typewriter)
Submitted By: @meisenzahl

### DIFF
--- a/applications/com.github.manexim.typewriter.json
+++ b/applications/com.github.manexim.typewriter.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/manexim/typewriter.git",
-  "commit": "c98fe94a62a65732aa6e4935782fb17aef90c37a",
-  "version": "0.6.0"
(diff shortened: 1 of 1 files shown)


danrabbit:
@meisenzahl can you provide some more information on why you've re-added accountsservice here?

meisenzahl:
@danrabbit of course. Thank you so much for your time!

An issue was submitted in https://github.com/elementary/appcenter/issues/1762 that some apps cannot be launched after installation from AppCenter and crash without error message. It looks like without applying the latest updates, running apps that have the sandbox hole for accountsservice removed are crashing.

It also seems that even after updating `io.elementary.Platform` the apps crash until a reboot.

Therefore, I have re-added accountsservice as a hotfix.

I suggest that we release a new version `6.1` of [flatpak-platform](https://github.com/elementary/flatpak-platform) so that this breaking change does not occur.

danrabbit:
Ah interesting case. Sounds like this is caused by the user holding back some updates but allowing others

PR #192, opened 2021-11-09
com.github.watsonprojects.docs 3.0.0

ChildishGiant:
This is the first release of [Docs](https://github.com/watsonprojects/docs), a rebrand of [quickDocs](https://github.com/mdh34/quickDocs). Currently not much has changed except the flatpaking, branding and support for dark theme but there are multiple PRs I'm going to merge once they've had a bit more work. 

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Name is unique and non-confusing
- [ ] Matches description
- [ ] Matches screenshot
- [ ] Launchable tag with matching ID
- [ ] Release tag with matching version and YYYY-MM-DD date
- [ ] Custom colors meet WCAG A contrast or greater
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime
- [ ] Sandbox permissions are reasonable

PR #206, opened 2021-11-19
com.github.watsonprojects.easydocs 3.0.1

ChildishGiant:
Replacing #192

## Review Checklist

- [x] App Opens
- [x] Does what it says
- [x] Categories match

## AppData

- [x] Matches Description
- [x] Matches Screenshot
- [x] OARS info matches

## Flatpak

- [x] Uses elementary runtime


danrabbit (changes requested):
@ChildishGiant only one issue here, the submission is for a version 3.0.1 but AppData only contains a release tag for 3.0.0 :)

ChildishGiant:
Any update with the review?

danrabbit (approved):
@ChildishGiant in the future, please avoid rewriting git history. This makes it so I have to redo the whole review from scratch instead of being able to review just the latest changes

PR #217, opened 2021-11-29
com.github.itmm.Contacts 1.2.2

itmm:
Would you be so kind to review my changes to the app Contacts. The original repository is on

[/aggalex/Contacts](/aggalex/Contacts)

but as the repository seems to be stale (no reply on issues, pull request and e-mail), I now try to add my fork to the App Center. I removed the warnings and added a flatpak buildfile.

### DIFF
--- a/applications/com.github.itmm.contacts.json
+++ b/applications/com.github.itmm.contacts.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/itmm/Contacts.git",
+  "commit": "a9b564e637ff97515209d2a0594709848c8892e1",
+  "version": "1.2.2"
+}


cassidyjames (commented):
Thanks for submitting this, @itmm! To clarify, are you committed to maintaining and further developing this fork? We don't generally allow forks of existing apps that are simply repackaging, but if you're developing an app based on an existing codebase, that could be treated differently.

cassidyjames (changes requested):
If the latter is the case, then you'll still need to make a number of changes before we can approve this:

- [ ] Unique branding to reduce confusion
   - [ ] A new unique app name in the .desktop, AppData, and app UI
   - [ ] New, unique icons
   - [ ] Corrected app author info in AppData https://github.com/itmm/Contacts/blob/a9b564e637ff97515209d2a0594709848c8892e1/data/com.github.itmm.contacts.appdata.xml.in#L6
   - [ ] Updated contact info in AppData https://github.com/itmm/Contacts/blob/a9b564e637ff97515209d2a0594709848c8892e1/data/com.github.itmm.contacts.appdata.xml.in#L60
- [ ] Updated screenshots using the elementary OS 6 defaults

Releases must also contain relevant release descriptions containing a user-facing summary of the changes: https://github.com/itmm/Contacts/blob/a9b564e637ff97515209d2a0594709848c8892e1/data/com.github.itmm.contacts.appdata.xml.in#L22

itmm:
@cassidyjames thanks for your reply.

>   * [ ]  A new unique app name in the .desktop, AppData, and app UI
>   * [ ]  New, unique icons

I would love to keep the old name and icon, as it basically is the original app made available for elementary OS 6. Should I change the id back to com.github.aggalex.Contacts to achieve that? I thought it is more clear, if the id points to the repository that is currently maintaining it. I am no good on graphics, so to create different artwork and name I first have to find collaborators.

danrabbit:
@itmm The app's name and icon are considered trademarks and aren't available under the same terms as the license for the code. You would need explicit permission from the original app author to use their trademarks

itmm:
Hmm, I guess that under this constraints it does not make sense for me to release a working version of Contacts. I wanted to tap into the efforts already done and allow to use the app in Elementary OS 6, not to build a new app.

PR #219, opened 2021-11-30
FlocoreClock v1.0.0 json

Michael-Curtis0:
FlocoreClock v1.0.0

### DIFF
--- a/applications/com.github.flocore.flocoreclock.json
+++ b/applications/com.github.flocore.flocoreclock.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/flocore/flocoreclock.git",
+  "commit": "6d980435954b914a86a342ecab20b653f153bf61",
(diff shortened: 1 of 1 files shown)


dar5hak:
@Michael-Curtis0 A couple of (unofficial) comments for the screenshot:
- Could you use the default elementary OS button layout (close on left, maximize on right)?
- Could you take the screenshot using the “Grab the current window” feature in the default Screenshot app? That way, you also capture the window shadows.

dar5hak:
You might also want to add the release information in the appdata: https://docs.elementary.io/develop/appcenter/publishing-updates#release-descriptions

dar5hak:
And it would be helpful to potential users if you could describe what features your app offers that the built-in clock in the panel doesn't.

danrabbit:
Closing since the developer hasn't responded since November

PR #225, opened 2021-12-02
org.small_tech.comet 1.1.1

aral:
Request to add Comet to the elementary OS AppCenter.

https://github.com/small-tech/comet

  - __Update:__ Version 1.1.0. No longer requires Flatpak filesystem permissions.
  - __Update:__ Version 1.1.1. Fix: add releases section to Flatpak metadata.

## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/org.small_tech.comet.json
+++ b/applications/org.small_tech.comet.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/small-tech/comet.git",
+  "commit": "2a75775c635da9e2a73f33d03d6926d0f3d43909",
+  "version": "1.0.0"
+}
--- a/applications/org.small_tech.comet.json
+++ b/applications/org.small_tech.comet.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/small-tech/comet.git",
-  "commit": "eab51817876cfa38a5295f92b4afd6f1aea9d31d",
-  "version": "1.0.0"
+  "commit": "3a38ecb25e54e060f98db226907d8048b6cef9dd",
+  "version": "1.1.0"
 }
--- a/applications/org.small_tech.comet.json
+++ b/applications/org.small_tech.comet.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/small-tech/comet.git",
-  "commit": "3a38ecb25e54e060f98db226907d8048b6cef9dd",
-  "version": "1.1.0"
+  "commit": "fdb8f07080b72e6b75e487ec5f24cb564fd3cfed",
+  "version": "1.1.1"
 }
--- a/applications/org.small_tech.comet.json
+++ b/applications/org.small_tech.comet.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/small-tech/comet.git",
-  "commit": "2a75775c635da9e2a73f33d03d6926d0f3d43909",
+  "commit": "e3a609ae23ca5356b9bb9c666508a90e600302ec",
   "version": "1.0.0"
 }
--- a/applications/org.small_tech.comet.json
+++ b/applications/org.small_tech.comet.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/small-tech/comet.git",
-  "commit": "e3a609ae23ca5356b9bb9c666508a90e600302ec",
+  "commit": "eab51817876cfa38a5295f92b4afd6f1aea9d31d",
   "version": "1.0.0"
 }


danrabbit:
Hey @aral I'm wondering where you got this Stripe key? If this is the public key for your stripe account, this isn't a valid key and payments will fail. Keys are generated that are specific to our stripe store and each individual app

aral:
@danrabbit I got it from the Stripe dashboard after creating and linking the account via Stripe Connect.

So this is the public/publishable key for our account.

I don’t know where to find the key that is specific for my app on the elementary OS Stripe store. Do you have a link to where that’s documented and/or where I can find it?

Does this mean that once I have that key, I should submit a different pull request?

Thanks!

danrabbit:
Yeah so we're in a bit of a limbo here since we don't have a connect integration yet built for the new dashboard and the old dashboard only spits out the keys as part of the debian packaging process

We did have one person who seems to have been able to successfully generate a new key, but nobody else has been able to reproduce this. From their description:

1. Go to the old AppCenter Dashboard and add your apps there and configure monetization on them
2. Go to your stripe Dashboard, open the Developer tab, and there should be a section marked something like “managed by elementary AppCenter”. This should have individual public keys for each of your apps

I haven't been able to find any documentation that backs this up or reproduces this, so I really have no idea how they did it :sweat_smile:

Yeah we'd need a new release tag either with a valid key or no key

aral:
Right, so here’s what I did:

1. I used the old/current dashboard to create my app:

    Screenshot from 2021-12-02 17-26-40@2x

2. During that process, I went through the Stripe Connect flow where I created a new account specifically for the elementary OS AppCenter and successfully linked that to whatever backend is being by the old/current dashboard (it shows are green in the interface, above and Stripe says the account is being managed by the elementary OS AppCenter)



The API Key is the one from there.

It is not the same as the API key for our Small Technology Foundation account, for example. But I also do not see any other keys there or any indication that it is specific to my app.

Thoughts?

Is there some way you can check on your end, through Stripe, whether that is the correct API key?

danrabbit:
@aral ah okay cool, if it's different from your small tech key, then it might be correct! Let's try publishing with this key.

I haven't been able to see a place anywhere on my end where I could see keys that have been connected to our account unfortunately.

So my only other concern here is the `host` permission is extremely broad in your manifest. Is there any way to scale that down at all? Or can you explain what your app is accessing outside of the user's home?

aral:
Cool, let’s give it a try. I mean worst case scenario, the payments will fail, right? (No biggie on my end; that’s not the reason I’m doing this.)

Re: permissions – the only thing the app accesses is the git commit message but that can be anywhere a git repository is. I think I broadened them when a commit from /tmp/ failed while I was running the screenshot script.

I can take another look if that’s an issue.

danrabbit:
@aral yeah if you could scale that down to `home` and `tmp` please. Flatpak automatically blacklists some dirs in host, but it's probably not reasonable to expect users to have git repositories outside of `home` and `tmp`

aral:
Right, just tried it with `home` and did some testing and remembered why I set it as `host` and not `home`: it cannot access external drives when set to `home`.

Screenshot from 2021-12-02 20-31-13@2x

danrabbit:
@aral In this case, I think you want to add `/run/media` or `/media` instead of blanket adding `host`. In general, we're trying to get filesystem permissions as restricted as possible. This documentation might be helpful: https://docs.flatpak.org/en/latest/sandbox-permissions.html

As an aside, you want to be really careful about using icons for things other than their semantic icon name, because metaphors can and do occasionally change. The more appropriate icon for this dialog is probably `dialog-error`

aral:
@danrabbit Hey Dani, having had a think about this (and re-read some of the information on Flatpak sandboxing that I hadn’t visited in a while) – and unless you still have objections after reading [the justification I just added to the source code](https://github.com/small-tech/comet/blob/main/org.small_tech.comet.yml#L17) – I’d like to keep `filesystem=host` permissions to ensure that the app works for projects at external mount points (which can be at a myriad of locations, not just `/media` in the wild):

> The app uses host permissions instead of home permissions to support projects
mounted as external drives and network shares. As mount points can (and do)
differ, it makes sense to use host permissions to ensure that all such projects
work instead of attempting to compile a list of all popular mount points to
support.
>
> In terms of security, what matters is whether an app has home permissions or not
since, once an app has home permissions, it can easily escape the sandbox and
run arbitrary code and even escalate its own Flatpak permissions (as the settings
file is kept within the person’s home directory).
>
> As the threat model here assumes that either that the app author is malicious
or that the app’s source repository has been compromised and modified by
a malicious actor, the use of granular filesystem permissions on an app that
already has home permissions is security theatre.
>
> In this scenario, the real security in the system is not provided by the
Flatpak sandbox but by the following two properties of the system:
> 
> 1. The app being free and open source.
> 
> 2. The extent to which the automated and manual reviews of the source code, as
    carried out by the elementary OS app review team, can catch malicious behaviour.
> 
> Sources:
> 
> - https://flatkill.org/2020/
> - https://github.com/flatpak/flatpak/issues/3637

So, all that to say that the actual security will be based on your review that the code doesn’t do anything malicious. In this case, the codebase is so small that you can easily verify that but, for larger projects, it might make sense to have some automated processes that use a virtual machine to check, for example, that shell startup scripts and other well-known sensitive locations within the home directory not being modified. (Again, this is not an alternative to a meticulous code review as such things can be done after X launches or X hours of use, etc., but I also realise that we are talking about a small team with limited resources carrying out these reviews while simultaneously maintaining the OS itself and working on the next version.) :)

Also, thanks for flagging the icon. I will look into that for the next update.

Is it better, would you think, if I implemented custom icons for such things if usage differs from the semantics of the standard ones?

Thanks for your time and suggestions + can’t wait to see Comet on the AppCenter :)

danrabbit:
@aral Maybe a better fix would then be to use the filechooser portal for users to select where their git project is stored? Again, ideally we want the sandbox to be as tight as possible and `host` is extremely broad. It's already been kind of a thing with apps requesting `home` where we're telling developers that they need to work on tightening up how their app works to fit into more restricted filesystem permissions.

I just realized that at the moment Comet is changing user settings on first run without asking permission, which is not something we want apps to do.

Flatkill is full of misinformation and not an authoritative source

danrabbit:
Ah yeah regarding icons, yes if you want to use an icon for a purpose other than its semantic name, I would recommend copying the icon into your app and loading it as a gresource under your app's namespace

aral:
Hey Dani,

Using a filechooser portal would not work as it would ruin the flow. In that case, basically any other app would have a better experience than Comet. Basically, there’d be no reason to have Comet; even nano would provide you with a better experience :)

About setting itself as your Git commit editor on launch: again, that’s about the experience. If you download Comet, you do so to test/use it as your Git commit editor. It has no other purpose. So it’s makes sense to activate it on launch (which is the gesture that means “I want to test/use it as my Git commit message editor.” So it saves you the effort of repeating yourself.

Also, it’s not destructive: it remembers your current setting so that if you want to go back to using your current editor, that’s just a click of a button to do.

The alternative is to have the person’s first experience on launch to be a negative one “Comet is not set as your Git commit message editor” and to expect the person to repeat themselves when they’ve already expressed their desire to use Comet as their Git commit message editor by downloading it and then launching it.

This is actually one of the experience improvements I implemented in Comet over the generic version of the tool I’d originally created from GNOME (and the fork of that by Sony, called Commit, is available on GNOME Spaces) and it’s one of the features I’ll be using to highlight how you can optimise the experience and reduce effort in your apps in my talks so I’d really like to keep those as they are.

(One of my goals with Comet is to highlight how you can optimise the experience when you focus on a single OS over targeting X number of systems by, saying, making a GNOME app.)

Regarding security: the most important consideration should be that you have the source, you know the app isn’t doing anything nefarious (and nor would it) so I really don’t want to impede the experience in any way.

Please let me know if any of these are a showstopper for getting Comet on the AppCenter and I’ll consider my options accordingly.

Thanks again for your time and feedback.

aral:
@danrabbit Hey Danielle, I’m sure you’re busy with all the other stuff you have to do so here’s a quick check list. Can you please let me know which of these two issues, if either, is a showstopper for getting Comet on the AppCenter as a developer app:

  - [ ] host permissions
  - [ ] setting itself as your Git editor on launch (with one-click ability to revert)

I’d really like to get the app on the AppCenter this week if possible as it’s holding up the rest of the work I want to do with [Watson](https://github.com/small-tech/watson) and the articles I want to write regarding the experience.

(I’d like to go back to focusing on building the small web before the end of the month so I’m just trying to get everything sorted before then.)

Hope you’re having a good week and please let me know what you need from my end to conclude this process :)

cassidyjames:
Hey @aral, I think I can chime in on your questions:

>host permissions

When this has come up on the Flatpak issue tracker, contributors have rightly pointed out that you can scope your permissions down, e.g. to standard mount points like `/mnt`, `/media`, and `/run/media` to cover most cases. In more obscure situations, users are still in control as they can override permissions by visiting System Settings → Applications → Permissions (which you can link to in your app with `settings://applications/permissions`).

>setting itself as your Git editor on launch (with one-click ability to revert)

As far as I understand, this would be considered in conflict with this section of the [AppCenter Publishing Requirements](https://docs.elementary.io/develop/appcenter/publishing-requirements#technical-requirements):

>Your app must not attempt to modify other apps or system programs.

However, I also understand that's not very convenient since it's kind of the point of your app. :smile: In this case, I think the _proper_ way forward would be to determine if there's a way for there to be a more sandbox-friendly way of changing this setting that keeps the user in control, e.g. via a Portal. However, I'm not sure how much traction you'd get there from other FreeDesktop contributors.

I'd have to confirm with @danrabbit, but I think at the very least to be accepted in AppCenter we would require you to prompt the user before making that change. I would recommend a simple MessageDialog that looks similar to what users would expect for other agents/permission requests. I know you've mentioned they have "consented" by installing the app, but I strongly disagree with that; installing or opening an app **should not** change system settings out from under users without explicit user consent, and that would be a terrible precedent to set with AppCenter apps.

I hope this helps and provides a path forward to getting you app on AppCenter!

danrabbit:
Hey sorry for taking so long to follow up! I've been in and out of work sick the past couple weeks unfortunately 🤧

Regarding changing the default editor for git, I feel like something to keep in mind is that there is no GUI way to change this if comet is uninstalled, leaving the user with a broken git configuration. So I would honestly recommend a welcome/placeholder/other onboarding that tells users they can set comet as their default editor with `git config`. This would make sure we have active and informed consent, wouldn't require a sandbox hole, and make sure users are able to change back the setting in case they decide to uninstall comet.

Cass and I discussed briefly on a recent call that we probably need to work on some documentation that's clearer regarding sandboxing requirements etc so sorry for the confusion and delay while we iron out these policies!

aral:
Hey @cassidyjames,

> Hey @aral, I think I can chime in on your questions:

Thanks :)

> > host permissions
> 
> When this has come up on the Flatpak issue tracker, contributors have rightly pointed out that you can scope your permissions down, e.g. to standard mount points like `/mnt`, `/media`, and `/run/media` to cover most cases. This is what popular apps like Kodi do, as well. In more obscure situations, users are still in control as they can override permissions by visiting System Settings → Applications → Permissions (which you can link to in your app with `settings://applications/permissions`).

The problem with this, as I was pointing out to @danrabbit earlier, is that once you have home (not host) permissions (which Comet needs and which, as a developer tool, makes sense) the difference between home and host permissions is security theatre. Once an app has home permissions, it is effectively not sandboxed because all it needs to do to have full user-level permissions to the system is to modify a launch-time script (e.g., ~/.bashrc, ~/.zshrc, etc.) to download and run arbitrary code.

So the security consideration here that matters is not whether or not Comet should be allowed to have host permissions but if it should be allowed to have home permissions. Once I have home permissions, if I am malicious, I can pretty much do whatever I want. Whether or not I am malicious and doing nasty things comes down the code review and to any automated tests you may or may not have that can detect such malicious behaviour (happy to chat with either of you separately to outline what those could look like as I feel you should definitely have them, possibly running under a VM, as an automated part of the review process).

So the file system security issue comes down to this:

Is Comet allowed to have file system access? (Home or host doesn’t matter.)

If not, and Comet must use a file portal to access files, then it doesn’t make sense to have on the AppCenter.

> > setting itself as your Git editor on launch (with one-click ability to revert)
> 
> I'd have to confirm with @danrabbit, but I think at the very least to be accepted in AppCenter we would require you to prompt the user before making that change. I would recommend a simple MessageDialog that looks similar to what users would expect for other agents/permission requests, or you could consider an infobar if that feels lighter-weight. I know you've mentioned they have "consented" by installing the app, but I strongly disagree with that; installing or opening an app **should not** change system settings out from under users without explicit user consent, and that would be a terrible precedent to set with AppCenter apps.

This is not a bad idea in any case. I’m not opposed to having a modal that prompts the person for permission and takes the opportunity to clearly state what is happening. The one extra click is worth having the person confident of what’s going on and in control.

> I hope this helps and provides a path forward to getting you app on AppCenter!

It think it does. I’m going to start working on the modal permission dialogue and, when you can, I’d appreciate it if you can confirm whether or not the remaining issue (file system access) is a showstopper for you (which I really don’t think it should be for a developer app that does what Comet does).

Thanks :)

aral:
> Hey sorry for taking so long to follow up! I've been in and out of work sick the past couple weeks unfortunately sneezing_face

Hey, no worries. Stuff can wait, please take care of yourself and rest if you need to.

I like to think we’re all in this for the long haul and that includes self-care :)

Hope you feel better soon.

aral:
@cassidyjames I’ve now updated the app to include the initial permission request dialogue. This now displays at launch if Comet is not set as your Git commit message editor asking you for permission to update your Git configuration for you:

Screenshot from 2021-12-09 14-53-47@2x

I’ll update the commit on this pull request once I have the translations for the new strings ready.

btkostner:
Hello, I just wanted to drop by and say that the Stripe token integration for the new AppCenter dashboard is live. You can now go to https://beta.developer.elementary.io/ and type your app URL in. There you will see a "Get a Stripe key" link to log in with Stripe. After you complete your on-boarding you will see the needed Stripe token. Let me know if you have any questions.

aral:
@btkostner Thanks for the update, Blake. I went through the flow and selected the new Stripe account I had created previously (when I ran through the old/current AppCenter process) and I simply got returned to the initial beta AppCenter page without any errors or warnings.

I don’t see any new keys.

Not sure whether I should delete that Stripe account and create a new one for the new flow.

Also, it seemed to only want to connect with my aral github account, not the Small Tech organisation account. Is that going to be a problem? Thanks!

cassidyjames:
>What I would suggest is that AppCenter allows developer tools that require home/host permissions but states that they are not sandboxed while listing them (because any Flatpak that has home permissions or above is not sandboxed).

Yes, this would be a feature request for AppCenter itself; probably solved by https://github.com/elementary/appcenter/issues/1012 or possibly reported as a more specific issue, and hanlded design-wise similar to how we handle exposing some OARS data up top.

>I’m happy either way (as I’m almost finished implementing a convenient publish task within Watson that outputs a web site complete with Flatpak repository with static deltas so folks can easily sideload Comet and get updates) 

That's dope! I think that speaks really positively to the overall benefits of elementary OS and our focusing on Sideload as a key piece of our platform. Even if your app is not accepted for whatever reason, it _should_ be easy to publish elsewhere like Flathub or even yourself.

Ultimately I'd love to get this sorted out to get Comet on AppCenter as well, but I do want to ensure we're setting the correct precedents and really thinking these cases through as we see them. I'm happy you have an alternative to get your app up and distributed in the meantime, too!

aral:
Cool. I’ll publish with my own repository then and update Watson’s workflows accordingly to make that the default. (And I look forward to hopefully having Comet on AppCenter once its policies are ready.) :)

aral:
@cassidyjames @danrabbit Hey Cassidy, Danielle,

I just opened an issue on the AppCenter repo for a security policy request on sandboxed and unsandboxed apps on AppCenter.

After a _very_ cursory ten-minute review, it seems that there is already precedent for having unsandboxed apps (apps with `--filesystem=home` and above Flatpak permissions) on AppCenter (see https://gist.github.com/aral/5b283d020bab878ca8b2cc83bc66b88d).

In the meanwhile, since the current (unofficial) policy is to accept unsandboxed apps (Inspektor, Akira,  EasySSH, Taxi, Trimir Journal, etc.), accepting Comet will not constitute a new precedent and I’d like to request that we go ahead with it unless the current thinking is that unsandboxed applications will not be allowed on AppCenter and the existing ones identified and removed.

In the meanwhile, I released Comet for sideload on https://comet.small-web.org – hope you enjoy using it and please do let me know if you notice any issues or have any feature requests :)

aral:
@danrabbit @cassidyjames I’ve just released Comet version 1.1.0 that removes the filesystem permission requirement, which was the last blocker on this pull request.

(Based on the technique that I saw Sonny using in Commit.)

Looking forward to hopefully seeing Comet on the AppCenter soon :)

danrabbit (approved):
Hey @aral,

Thanks for working with us on the filesystem permissions! This is much better. I think I'd still rather not see a command line here to change users' defaults, but since you're asking explicit permission here, I'm not gonna hold you up on it.

I have some additional, non-blocking comments:

Would you be able to leave a comment about your use case here for asking permission to be the default git commit message editor? That would help us to pursue this in a more sandboxed way: https://github.com/elementary/portals/issues/6

I would highly recommend against using "Yes" and "No" as dialog buttons. Button labels should be actions like "Set as Default" so that it's super explicit what will happen

Accountsservice permissions are no longer necessary for dark style. This is now exposed over the settings portal, so you can drop that hole from your sandbox in a future version as well

There's also a new platform version `6.1` available and we recommend upgrading from `6` in your manifest

aral:
@danrabbit Only just saw this. Will keep these in mind as I continue to update the app, thank you :) (And Happy New Year!)

PR #232, opened 2021-12-05
com.github.alexkdeveloper.tictactoy

alexkdeveloper:
## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [ ] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.alexkdeveloper.raddiola.json
+++ b/applications/com.github.alexkdeveloper.raddiola.json
@@ -1,5 +1,0 @@
-{
-  "source": "https://github.com/alexkdeveloper/raddiola.git",
-  "commit": "d24d716637c27961cbbab29523d96090c4e8762a",
-  "version": "1.0.0"
-}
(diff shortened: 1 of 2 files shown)


danrabbit (approved):
Hey @alexkdeveloper this looks great! I have a few comments, but none of them are blocking :)

A new version of our Flatpak runtime `6.1` is available and we recommend you upgrade to that from `6` in your Flatpak manifest

In your desktop file, I think `LogicGame` could be added to your categories at the least. Probably worth looking through the additional categories section of the spec to see if others apply: https://specifications.freedesktop.org/menu-spec/latest/apas02.html

Also, if you decided you want to monetize your app, we're now generating new Stripe keys at https://beta.developer.elementary.io/

Thanks again for your submission!

PR #235, opened 2021-12-07
Add applications/com.github.jeremypw.gnonograms.json

jeremypw:
Repository: [ jeremypw/gnonograms](https://github.com/jeremypw/gnonograms)
Submitted by: @jeremypw

 - First attempt at Flatpaking Gnonograms
 - Use new app icon

## Review Checklist
- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.jeremypw.gnonograms.json
+++ b/applications/com.github.jeremypw.gnonograms.json
@@ -1,0 +1,5 @@
+{
+   "source": "https://github.com/jeremypw/gnonograms.git",
+   "commit": "c959a30e8a761cb309e66eded5a061db83836f9b",
+   "version": "2.0.0"
+ }


danrabbit (approved):
Hey @jeremypw nice work updating your app! I have a few comments, but they aren't blocking :)

The new `6.1` runtime is out, so I recommend updating from `6` in your manifest to the latest version

In your desktop file, I think `LogicGame` could be added to your categories at the least. Probably worth looking through the additional categories section of the spec to see if others apply: https://specifications.freedesktop.org/menu-spec/latest/apas02.html

Also, the sandbox hole for `accountsservice` is no longer required in your manifest. There's a settings portal for the dark style now, so you can remove this from your finish args and it'll still work

I'm curious also about the sandbox hold for xdg-documents. I think the document portal should persist access to anything you open/save with the file chooser dialog right?

danrabbit:
Oh one more thing! If you decided you want to monetize your app, we're now generating new Stripe keys at https://beta.developer.elementary.io/

jeremypw:
@danrabbit  Thanks for the review and suggestions!  I'll update the manifest and desktop file and check everything still works.

jeremypw:
@danrabbit I tried removing the xdg-documents hole but then weird things happened saving and reloading a game.  Saving appeared not to save the game to the true `~/Documents` but to under`/run` (as reported by Gnonograms).  Also, trying to reload the game then showed a weird name in `~/Documents` with extra characters added before and after the name it was saved with. This mangled file didin fact exist in the true `~/Documents`folder` (not under `/run`) and contained the game saved.  So the main issues are (1) Gnonograms reports that the file was saved under `/run/usr/1000/app/com.github/jeremypw/gnonograms` and (2) the file that it was saved under had a mangled name.  

With the xdg-documents hole then the expected behaviour occurs - i.e. Gnonograms reports that the file was saved under `~/Documents` and there is a file there with the expected name.

Any idea what I need to do to make use of the documents portal transparent?

danrabbit:
@jeremypw thanks for taking the time to look into it! Honesty I'm not 100% sure. That's something I feel like I need to set aside time for is really dive into using the document portal and file chooser portal and then write good developer docs about saving and loading files

jeremypw:
@danrabbit Yes, me too!  I am still pretty hazy about many aspects of Flatpak.  Another I am not sure about is whether the "Categories" section need/should contain "Gnome;GTK".  The only directly relevant categories seem to be "Game;LogicGame".

danrabbit:
@jeremypw yeah we don't do anything with those categories in AppCenter currently. It might be useful in the future for 3rd party sources maybe to try to like indicate some kind of compatibility spectrum? It's possible KDE Discover or GNOME Software do something with those in their stores?

jeremypw:
For now I am going to leave the "Gnome;GTK" categories as they do not do any harm and also leave the "xdg-documents" filesystem hole until the document portal is added to `elementary/portals` and its use made easier.  There does not seem to be any Vala code on Github making use of it yet. I do not yet see how it works with the filechooser portal (if it does).

PR #237, opened 2021-12-09
TenderOwl/Frog 0.2.0 Release

amka:
Repository: [tenderowl/frog](/tenderowl/frog)
Submitted By: @amka

### DIFF
--- a/applications/com.github.tenderowl.frog.json
+++ b/applications/com.github.tenderowl.frog.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/TenderOwl/Frog.git",
-  "commit": "06346c027eb9dffed5866d0157c34525261708de",
-  "version": "0.1.7"
+  "commit": "24a7672ba0f27f6b3bf28788c4b9435c88ef5fda",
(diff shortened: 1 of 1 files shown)


danrabbit:
Nice update @amka!

While you're here, I wanted to let you know that a new version of our Flatpak runtime `6.1` is available and we recommend you upgrade to that from `6` in your Flatpak manifest :)

Also, the sandbox hole for `accountsservice` is no longer required. There's a settings portal for the dark style now, so you can remove this from your finish args and it'll still work

Finally, if you'd like to monetize your app, we're now generating new Stripe keys at https://beta.developer.elementary.io/

danrabbit:
Ah, actually one more thing. I noticed that in your appdata, your release tag is missing a `date` and instead uses `timestamp`. According to the [appstream docs](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases) `timestamp` should no longer be used. I think theoretically it still should work for now, but it may not work in future versions so I would recommend switching that before your next release.

Thanks!

danrabbit (approved):

PR #239, opened 2021-12-10
com.github.avojak.iridium 1.7.0

avojak:
1.7.0 release of Iridium: https://github.com/avojak/iridium/releases/tag/1.7.0

### DIFF
--- a/applications/com.github.avojak.iridium.json
+++ b/applications/com.github.avojak.iridium.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/avojak/iridium.git",
-  "commit": "6dd49583a001174c8cf902c7251931d010679040",
-  "version": "1.6.1"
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Hey @avojak great update!

I have a couple suggestions but nothing blocking your update :)

A new version of our Flatpak runtime `6.1` is available and we recommend you upgrade to that from `6` in your Flatpak manifest :)

When adding custom icons, I would highly recommend using a gresource. We have some documentation about gresourcing icons [here](https://docs.elementary.io/develop/apis/gresource#icons). Let me know if any part of that is confusing or insufficient and I can follow up with you there to address those issues

Also, the sandbox hole for accountsservice is no longer required. There's a settings portal for the dark style now, so you can remove this from your finish args and it'll still work

Thanks!

avojak:
@danrabbit Thanks for the suggestions! I'll be sure to get those addressed for the next release!

PR #241, opened 2021-12-10
com.github.brain child.moobo 0.1.2

pierrefabarius:


danrabbit (approved):
Hey @pierrefabarius nice update! I have a couple of comments, but they aren't blocking :)

A new version of our Flatpak runtime `6.1` is available and we recommend you upgrade to that from `6` in your Flatpak manifest 

I wonder if `ProjectManagement` would be a good addition to your Desktop Entry categories? Might be worth looking over the extended categories here: https://specifications.freedesktop.org/menu-spec/latest/apas02.html

PR #248, opened 2021-12-15
com.github.gabutakut.gabutdm 1.6.0

torikulhabib:
Add style.
Window.
Open link 🔗.
Icon generic to icon.
Notify icon.

### DIFF
--- a/applications/com.github.gabutakut.gabutdm.json
+++ b/applications/com.github.gabutakut.gabutdm.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/gabutakut/gabutdm.git",
-  "commit": "b6c85c0c105014ff04dfdb2e08b1a019406253ea",
-  "version": "1.6.0"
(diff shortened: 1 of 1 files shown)


danrabbit (changes requested):
Hey @torikulhabib thanks for submitting this update!

Your app shouldn't be accessing accountsservice here. The dark style preference is now exposed over the settings portal. You can either use [Granite.Settings](https://valadoc.org/granite/Granite.Settings.html) for an easy way to get this, or if you don't want to use Granite, you can find a sample of how to call the settings portal directly [here](https://github.com/elementary/granite/blob/912ca324183ae14feb2ca96e0f09cbe68423eb9d/lib/Widgets/Settings.vala#L99)

torikulhabib:
https://github.com/gabutakut/gabutdm/commit/c9ebcab2a2042931b03784ee6635c4dd5266491c

danrabbit (approved):

PR #252, opened 2021-12-17
com.github.ryonakano.pinit 1.3.0

ryonakano:
Repository: [ryonakano/pinit](https://github.com/ryonakano/pinit)
Submitted By: @ryonakano

## Review Checklist
  
  - [ ] App opens
  - [ ] Does what it says
  - [ ] Categories match

  ### AppData
  - [ ] Matches description
  - [ ] Matches screenshot
  - [ ] OARS info matches
  
  ### Flatpak
  - [ ] Uses elementary runtime

### DIFF
--- a/applications/com.github.ryonakano.pinit.json
+++ b/applications/com.github.ryonakano.pinit.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ryonakano/pinit.git",
-  "commit": "aa09befa856f92640649bfa069694d3ce7232581",
-  "version": "1.2.1"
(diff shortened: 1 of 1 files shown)


danrabbit (approved):

danrabbit:
Nice work! I have a couple non-blocking comments:

There's a new flatpak runtime available, version `6.1` so I recommend updating that from `6` in your Flatpak manifest

is it possible to scope down the `  - '--filesystem=home'` permission to just `~/.local/share/applications`?

ryonakano:
Thank you for your suggestions! I addressed these changes so it should be fixed in the next release.

PR #253, opened 2021-12-17
com.github.matthiasjg.trimirjournal 1.0.6

matthiasjg:
Kindly asking for review and publish of bug fix release (a regression of removing Flatpak filesystem permissions preventing reset/ restore and backup of journal).

Also, I wanted to ask when the app will be listed on the https://appcenter.elementary.io website—or if at all?
In this regard I noticed, that the app's website is somehow broken: https://appcenter.elementary.io/com.github.matthiasjg.trimirjournal/

Is this because I do something wrong, maybe in `appdata.xml`? If so a pointer would be much appreciated. Thank you.


danrabbit (approved):

danrabbit:
Hm, can you open an issue report about that at https://github.com/elementary/appcenter-web/issues ? Probably would be the best place to get the attention of the right people to look into that for you!

PR #258, opened 2021-12-18
com.github.donadigo.appeditor 1.1.2

donadigo:
Repository: [donadigo/appeditor](https://github.com/donadigo/appeditor)
Submitted By: @donadigo
## Review Checklist
* [ ] App opens
* [ ] Does what it says
* [x] Categories match

### AppData
* [x] Name is unique and non-confusing
* [x] Matches description
* [x] https://github.com/donadigo/appeditor/issues/117
* [ ] Matches screenshot
* [x] Custom colors meet WCAG A contrast or greater
* [x] OARS info matches

### Flatpak
* [x] Uses elementary runtime
* [ ] Sandbox permissions are reasonable


danrabbit:
Can you explain why you need `"--socket=system-bus"`?

This isn't blocking, but there's a `6.1` runtime now, so I'd recommend upgrading from `6` in your manifest before your next release

donadigo:
I need `--system-bus` because `Granite.Settings` needs access to the system bus when setting up the color scheme: https://github.com/elementary/granite/blob/9271fa28c2f72c33020564ceec2d4339f977789d/lib/Widgets/Settings.vala#L120. Otherwise the app crashes on that step.

Screenshots updated and new release made. Thank you for a quick review!

danrabbit:
@donadigo ah no that's only a fallback for older apps that are using accountsservice. Since a couple of Granite versions, the dark style preference is accessible through the settings portal. But even if for some reason neither of those are available, it should still fall back instead of crashing. If you're getting a crash, you might have an old version of the 6 runtime. Updating to 6.1 would force a newer granite

danrabbit (changes requested):
Your app should work with the dark style through the settings portal, especially now that it's using the 6.1 runtime. If for some reason you're still getting a crash, you can add `--system-talk-name=org.freedesktop.Accounts` and that should resolve it. But full on system bus access is too much :sweat_smile:

donadigo:
@danrabbit Updated to 6.1. Works nicely without `--system-bus` and with no additional permissions. Thank you!

danrabbit (approved):
Fantastic! Nice job being app #100 ;)

PR #250, opened 2021-12-16
hr.from.josipantolis.starfish 1.3.0

Antolius:
Repository: [starfish-app/Starfish](https://github.com/starfish-app/Starfish)
Changes: [1.2.0...1.3.0](https://github.com/starfish-app/Starfish/compare/1.2.0...1.3.0)
Submitted By: @Antolius

PR #259, opened 2021-12-18
hr.from.josipantolis.starfish 1.3.1

Antolius:
Repository: [starfish-app/Starfish](https://github.com/starfish-app/Starfish)
Changes: [1.3.0...1.3.1](https://github.com/starfish-app/Starfish/compare/1.3.0...1.3.1)
Submitted By: @Antolius

### DIFF
--- a/applications/hr.from.josipantolis.starfish.json
+++ b/applications/hr.from.josipantolis.starfish.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/starfish-app/Starfish.git",
-  "commit": "aded6e68dba5c771129f0b6d89d23a98546a89fb",
-  "version": "1.3.0"
+  "commit": "c67d3efd7867cd3889df337a015d8e49ab6b4b44",
+  "version": "1.3.1"
 }


Antolius:
As suggested by @danrabbit in [a comment](https://github.com/elementary/appcenter-reviews/pull/250#pullrequestreview-835552723) on the recent 1.3.0 pull request, I've upgraded to the `6.1` runtime version, and removed the `org.freedesktop.Accounts` permission. While testing the new runtime I also discovered a few bugs which I've fixed with this release.

I've also noticed an issue with my `appdata` screenshot links. I've previously linked to files on the default branch of my GitHub repo, something I believe is a common practice for AppCenter apps. However, that creates an inconsistency in period between new release, which is typically created on the default branch, and publishing the update in this repo. During this period AppCenter will contain listing for the old version, but default branch will have screenshots of the new one. This will either break the screenshots (in case their filenames changed) or will miss-represent the app (in case filenames remained the same and just their content changed). To solve this I've switched to linking to files on the specific git tag corresponding to the release, instead of the default branch. It might be a good idea to check for this in your standard "Matches screenshot" review step for other apps as well.

danrabbit (approved):

PR #260, opened 2021-12-19
Com.GitHub.alexkdeveloper.tictactoy 1.0.1

alexkdeveloper:


danrabbit (changes requested):
Hey @alexkdeveloper please update the existing file for tictactoy rather than creating a new one

alexkdeveloper:
I'm having trouble updating. I don't know why. The file on the old request has been deleted, but a new one is not being created.

alexkdeveloper:
I tried in different ways and nothing worked out. The file cannot be edited. Maybe I'm doing something wrong)

PR #285, opened 2022-01-03
Update Eddy

donadigo:
Repository: [donadigo/eddy](https://github.com/donadigo/eddy)
Submitted By: @donadigo
## Review Checklist
* [ ] App opens
* [ ] Does what it says
* [ ] Categories match

### AppData
* [ ] Name is unique and non-confusing
* [ ] Matches description
* [ ] Matches screenshot
* [ ] Custom colors meet WCAG A contrast or greater
* [ ] OARS info matches

### Flatpak
* [ ] Uses elementary runtime
* [ ] Sandbox permissions are reasonable

### DIFF
--- a/applications/com.github.donadigo.eddy.json
+++ b/applications/com.github.donadigo.eddy.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/donadigo/eddy.git",
-  "commit": "b32868c4c10de67aee9ddcc2c511f911382d512d",
-  "version": "1.3.1"
+  "commit": "43a53fdd423e0d342e41b59f9d61ac157e80f088",
+  "version": "1.3.2"
 }


danrabbit (approved):
Simple enough update! Not gonna hold you up since this isn't anything new, but some thing I'd like to see in future updates:

There's a new runtime `6.1`. so you'll want to update from `6` in your flatpak manifest

Also, you should no longer need the sandbox hole for `org.freedesktop.Accounts`

If possible, I'd also like to see this `--socket=system-bus` dropped. If something in PackageKit is using it, maybe we can find a more specific `--system-talk-name`

Would it also be possible to change that `home:ro` access? Is that for scanning the downloads folder? If so can we scope it down to that?

Overall, it looks like we approved this with some pretty wide permissions. I know you're not doing anything too crazy, but the plan is to start warning users about these things in AppCenter, so we want to make sure your app doesn't look like it's requesting more permissions than it needs

PR #291, opened 2022-01-08
Added com.github.rajsolai.response v1.1.1

RajSolai:
Requesting Review for com.github.rajsolai.response 

## Review Checklist
- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.rajsolai.response.json
+++ b/applications/com.github.rajsolai.response.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/RajSolai/Response.git",
+  "commit": "00db6cd0cd040c446bc53caa97307718bf89c5a5",
+  "version": "1.1.1"
+}


danrabbit (approved):
Looks good nice work! :tada: 

A couple non-blocking comments:

A new version of the elementary runtime is out, so you'll want to update from `6` to `6.1` in your manifest in your next release

You only need to list OARS tags for which the value is not `none`. So if you don't have any content warnings to disclose, you can simply self close the OARS tag like this: `<content_rating type="oars-1.1" />`

I would recommend including additional categories in your desktop entry from [this spec](https://specifications.freedesktop.org/menu-spec/latest/apas02.html). At least `WebDevelopment` seems to apply here. This will help your app remain discoverable as the store grows

RajSolai:
Thanks a Lot, I will check into the categories in desktop entry and OARS guidelines

PR #294, opened 2022-01-09
Transfer existing app from GitHub RDNN to own domain

### LINKED ISSUE #293: Transfer existing app from GitHub RDNN to own domain
### Problem

Until recently, only apps with an RDNN were accepted according to the following scheme: `com.github.<user>.<repository>`.

However, it should be possible to transfer an already published app to its own domain.

### Proposal

It should be possible to specify the new RDNN with the entry `end-of-life-rebase` when a reason for `end-of-life` is specified.

## Flatpak

> hm, so instead of removing the old file, we need to rebuild that package with `end-of-life-rebase` as an argument to flatpak builder: https://docs.flatpak.org/en/latest/flatpak-command-reference.html?highlight=end-of-life-rebase
> 
> I think you need to file a new feature request against this repo where you can add `end-of-life` and `end-of-life` properties to the json file. And I'm also not sure if the Flatpak builder action supports these parameters, so we might need an upstream feature request there as well

Posted by @danrabbit in https://github.com/elementary/appcenter-reviews/pull/292#issuecomment-1008096015

## Flathub

> There may come a point where an application is no longer maintained. In order to inform users at update or install time that it will no longer get updates, create flathub.json with these contents:
>
> ```json
> {
>   "end-of-life": "This application is no longer maintained because..."
> }
> ```
>If the application has been renamed, you can additionally include end-of-life-rebase with the new ID. Recent flatpak versions will prompt user if they'd like to switch to the renamed app.
>
> ```json
> {
>   "end-of-life": "The application has been renamed to...",
>   "end-of-life-rebase": "the.new.appid"
> }
> ```

Source: https://github.com/flathub/flathub/wiki/App-Maintenance#end-of-life

meisenzahl:
Closes #293﻿

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,6 +19,8 @@ jobs:
       version: ${{ steps.parse.outputs.version }}
       source: ${{ steps.parse.outputs.source }}
       commit: ${{ steps.parse.outputs.commit }}
+      end-of-life: ${{ steps.parse.outputs.end-of-life }}
+      end-of-life-rebase: ${{ steps.parse.outputs.end-of-life-rebase }}
 
     steps:
       - name: Checkout
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -79,8 +79,10 @@ jobs:
           arch: ${{ matrix.arch }}
 
       - name: Deploy
-        uses: bilelmoussaoui/flatpak-github-actions/flat-manager@v3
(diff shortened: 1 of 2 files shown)


btkostner (commented):
You need to upgrade `elementary/action-appcenter-review-parse` to `elementary/action-appcenter-review-parse@v1.1.0`.

Also, not really a big thing, but you shouldn't need to add those inputs to `review.yml` because they are not used.

btkostner (approved):
This looks good

PR #295, opened 2022-01-10
com.github.avojak.warble 1.0.1

avojak:
First submission of new app: [avojak/Warble 1.0.1](https://github.com/avojak/warble/releases/tag/1.0.1)

## Review Checklist
- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.avojak.warble.json
+++ b/applications/com.github.avojak.warble.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/avojak/warble.git",
+  "commit": "462ae10f291c766a4a094662ab809a287e4ef3b6",
+  "version": "1.0.1"
+}


danrabbit (approved):
Great job! Really well done!

A couple non-blocking comments:

There's a new flatpak runtime out, so you'll want to update from `6` to `6.1` in your flatpak manifest before your next release

Your flatpak manifest has a sandbox hole for network, but I don't really see where that's being used. Possibly a mistake?

You only need to include OARS tags who have a value other than `none`. So if you don't have any content warnings to disclose, you can self close the tag like `<content_rating type="oars-1.1" />`

avojak:
@danrabbit Those first two points are due to bad copypasta from my outdated app template... 🤦🏼‍♂️ Good to know about the OARS metadata, that'll certainly clean the app data up a bit! I'll have all three items addressed for the next release - thanks so much!

PR #296, opened 2022-01-11
com.github.ryonakano.reco 3.2.1

ryonakano:
Changes: https://github.com/ryonakano/reco/compare/3.2.0...3.2.1


danrabbit:
Instead of doing a compile-time option for Pantheon, you should be able to use `GLib.Environment.get_variable ("XDG_CURRENT_DESKTOP");` to read your desktop environment on runtime. This would be better in case someone is running Pantheon on another distro or in case someone installs Reco from Flathub on elementary OS for some reason

danrabbit (approved):

PR #297, opened 2022-01-11
com.github.skarva.lockbox 1.3.1

danrabbit:
https://github.com/skarva/lockbox

## Review Checklist

- [x] App opens
- [x] Does what it says
- [ ] Categories match

### AppData
- [x] Matches description
- [x] https://github.com/skarva/lockbox/issues/26
- [x] Matches screenshot
- [x] https://github.com/skarva/lockbox/issues/27
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.skarva.lockbox.json
+++ b/applications/com.github.skarva.lockbox.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/skarva/lockbox.git",
+  "commit": "7c86b9d01af4fb2891bb985d9f184168de13e080",
+  "version": "1.3.0"
+}
--- a/applications/com.github.skarva.lockbox.json
+++ b/applications/com.github.skarva.lockbox.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/skarva/lockbox.git",
-  "commit": "7c86b9d01af4fb2891bb985d9f184168de13e080",
-  "version": "1.3.0"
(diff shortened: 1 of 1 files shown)


danrabbit:
Hey @zeebok! Filed a couple of issue reports about things missing to be published. Couple of other non-blocking comments:

You only need to include tags in OARS when the value isn't `none`. If you don't have any content warnings to disclose, you can self close the OARS tag like: `<content_rating type="oars-1.1" />`

I noticed there's a suggested price in appdata, but no stripe key. If you'd like to monetize, you can generate a new stripe key at https://developer.elementary.io and include that in appdata in a `x-appcenter-stripe` custom tag

zeebok:
Just made a new that addresses everything

zeebok:
Oh I didn't notice the "Categories match" item. Is that in reference to the desktop file? I just went off of what the metainfo creator recommended and generated.

danrabbit:
@zeebok Gonna review the changes now!

Regarding categories, yes that's with the desktop file. I would recommend adding `Security` from https://specifications.freedesktop.org/menu-spec/latest/apas02.html. We're considering removing the `Utility` category from the homepage in future versions of AppCenter since it ends up as kind of a junk drawer

danrabbit:
Update looks good! In the next release please match the exact version from your appdata and git tags. So `1.3.1` to `1.3.1`, not `1.3`

PR #300, opened 2022-01-14
com.github.eksanos.eksanos

jlahman:
Eksanos v1.0.1

## Review Checklist
- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.eksanos.eksanos.json
+++ b/applications/com.github.eksanos.eksanos.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/eksanos/eksanos.git",
   "commit": "cbb0086dc1a34921921b199a9f193c5a836f90f9",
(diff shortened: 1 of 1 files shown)


danrabbit (changes requested):
Hey thanks for submitting your app!

Everything looks except that your metainfo file doesn't contain any release tags. You'll need to add a release tag in your metainfo file and then tag a new release in GitHub so we can get this published. More info in the docs here: https://docs.elementary.io/develop/appcenter/publishing-updates#release-descriptions

danrabbit (approved):
Looks good, nice work!

PR #301, opened 2022-01-16
Resizer 2.2.0

peteruithoven:
Fixes: https://github.com/peteruithoven/resizer/issues/79

### DIFF
--- a/applications/com.github.peteruithoven.resizer.json
+++ b/applications/com.github.peteruithoven.resizer.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/peteruithoven/resizer.git",
-  "commit": "596a8f8e58f535b46b7f7645136fccabc9d67be3",
-  "version": "2.0.3"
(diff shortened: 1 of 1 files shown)


peteruithoven:
```
Error: HttpError: Resource not accessible by integration
Error: Resource not accessible by integration
```
No idea... What resource?

danrabbit (approved):
Nice update!

Some non-blocking comments for your next version:

There's a new Flatpak platform available, so you'll want to update from `6` to `6.1` in your manifest

I'm assuming the accountsservice hole was added for the dark style, but we use the settings portal for that in Granite now so you can remove that from your manifest

Is it possible to scope down your app's permissions from `home`? It looks like you're using the native file chooser which doesn't require filesystem permissions

peteruithoven:
Thanks for the review Danielle!

peteruithoven:
> Is it possible to scope down your app's permissions from home? It looks like you're using the native file chooser which doesn't require filesystem permissions

The app also supports drag & drop, which does seem to require the home filesystem permission. (I tried)

PR #307, opened 2022-01-24
Update to Time Limit v2.1.0

marbetschar:
Just a small maintenance release. Notable changes:

- Updated the Stripe key according to the new docs in the developer dashboard
- Update to elementary Platform v6.1
- Fixed app categories
- Fixed Portuguese and Brazilian Portuguese translations


danrabbit (approved):
Nice little update!

Since Granite uses the settings portal for the dark style, you should no longer need `accountsservice` in your sandbox holes, so don't forget to drop that before your next release.

I'm also wondering why you have `gnome-settings-daemon` here?

PR #310, opened 2022-01-30
Added com.github.rajsolai.textsnatcher v1.0.0

RajSolai:
TextSnatcher, OCR application

## Review Checklist
- [x] App opens
- [x] Does what it says
- [x] https://github.com/RajSolai/TextSnatcher/issues/3
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime
- [x] https://github.com/RajSolai/TextSnatcher/issues/1
- [x] https://github.com/RajSolai/TextSnatcher/issues/2

### DIFF
--- a/applications/com.github.alexkdeveloper.bmi.json
+++ b/applications/com.github.alexkdeveloper.bmi.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/alexkdeveloper/bmi.git",
-  "commit": "7af880cc89acb962df37c3ab3af0331b7561c897",
-  "version": "1.0.1"
+  "commit": "00a36158807e24799b4462546d291ba7cd2fffd3",
+  "version": "1.0.2"
 }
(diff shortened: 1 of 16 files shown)


danrabbit (changes requested):
Hey thanks for your submission! I filed a few blockers as issue reports on your GitHub repo.

I have one non-blocking comment as well: there's a new release of the elementary Flatpak runtime available, so you'll want to update from `6` to `6.1` in your manifest

RajSolai:
Thanks a lot @danrabbit :)

danrabbit (changes requested):
Hey @RajSolai 

It looks like you added `"--system-talk-name=org.freedesktop.Accounts"` to your sandbox holes. Your app shouldn't need access to that, can you explain why you added it?

I'm also wondering about the network access. As far as I can tell your app shouldn't need network access either

RajSolai:
- [--system-talk-name=org.freedesktop.Accounts](https://valadoc.org/granite/Granite.Settings.prefers_color_scheme.html) Granite Needs this permission for prefers_colors_scheme method but for elementaryOS it seems there is no need for this permission.
- Network permission was due to the template I used sorry.
I will remove the permission and commit the changes, Thank you @danrabbit

danrabbit (approved):
Looks good! Thanks for tightening up those permissions :)

PR #315, opened 2022-02-03
Add com.github.erkielbrecht.sticky-0.0.3

erkielbrecht:
## Review Checklist
- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.erkielbrecht.sticky.json
+++ b/applications/com.github.erkielbrecht.sticky.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/erkielbrecht/sticky.git",
-  "commit": "2676839ff2db29e509a58455733672a345eb4c48",
-  "version": "0.0.2"
+  "commit": "1ff4b6294f85d79dadebb6e57ccb99257555e1e8",
+  "version": "0.0.3"
 }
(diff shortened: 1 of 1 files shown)


danrabbit (changes requested):
Hey thanks for your app submission! This looks really good, I just have a couple of things here that need changes before we can publish:

The category to use in your desktop launcher should probably be `Office`. `Utility` is okay, but I general discourage people using that if there are more appropriate categories since it tends to be kind of a junk drawer. `System` is probably not accurate here.

The custom colors that you're using in your appdata do not meet WCAG AA color contrast levels. 

Screenshot from 2022-02-03 15 43 35

It looks like you set a suggested price, but your appdata doesn't contain your stripe key. If you'd like to monetize, you can generate a key here: https://developer.elementary.io/auth/github

We also have a new version of the elementary flatpak platform, so you'll want to update from `6` to `6.1` in your manifest

erkielbrecht:
Hey! Thanks for your suggestions. I tried to address them all in the latest release.

danrabbit (approved):
Everything looks good! Thanks for your patience :tada:

PR #319, opened 2022-02-08
Update com.github.alexkdeveloper.guessnumber 1.1.0

alexkdeveloper:


danrabbit (approved):
Looks good, thanks for your patience!

Just a note, I would recommend using the `destructive-action` and `suggested-action` style classes for your colored buttons instead of custom CSS here. You could also use a custom css snippet with `@define-color accent_color green` for example combined with `suggested-action`. This would fix the contrast levels for these buttons. Right now, the labels don't pass WCAG AA contrast level

PR #320, opened 2022-02-09
Add com.github.alainm23.planner

alainm23:
## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.alainm23.planner.json
+++ b/applications/com.github.alainm23.planner.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/alainm23/planner.git",
-  "commit": "67c86d0262bd9c7a560afbd32091afa166638019",
-  "version": "3.0.6"
+  "commit": "b0ee84d9e77cdb18026fdab8c032945b0fed4504",
+  "version": "3.0.7"
 }
--- a/applications/com.github.alainm23.planner.json
+++ b/applications/com.github.alainm23.planner.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/alainm23/planner.git",
-  "commit": "79eb825d91ba01bd44dcf3641877a8a1cc3b2b12",
-  "version": "3.0.1"
+  "commit": "5a635cbed659abf23d96241312899cd04d6087c3",
+  "version": "3.0.2"
 }
--- a/applications/com.github.alainm23.planner.json
+++ b/applications/com.github.alainm23.planner.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/alainm23/planner.git",
+  "commit": "79eb825d91ba01bd44dcf3641877a8a1cc3b2b12",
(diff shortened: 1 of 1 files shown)


danrabbit (commented):
Thanks for your patience! This looks really good as always. I have just a couple of questions:

Is it possible to scope down that `home` filesystem permission to more specific directories? In general I'm really trying to avoid wide permissions like this and we'll begin warning about them in AppCenter in the future

You have `social-chat` and `social-audio` marked as "moderate" in your OARS data. Is this right? This refers specifically to moderated communication between different users, like a forum for example or a moderated chat room.

And a couple of non-blocking comments:

Just so you know you don't need to include any OARS tags which have a value of `none`. Those are implicitly understood to be `none` by default :)

There's a new version of the Flatpak platform available, so you'll want to update from `6` to `6.1` in your manifest

alainm23:
@danrabbit Thanks for waiting, I am sending a new version 3.0.7.

The `home` folder permissions are necessary so that a user who was using Planner 2 can select the destination of his backup jobs and import them from Planner 3.

I have removed the OARS tags that Planner does not use.

danrabbit:
@alainm23 hm if this user is making a selection it sounds like this could be done with the file chooser portal no?

alainm23:
If there is such a possibility it would be amazing, do you have an example in Vala?

danrabbit:
It looks like this latest release doesn't build:
```
Error: Build failed: YAMLException: bad indentation of a mapping entry at line 87, column 18:
[20](https://github.com/elementary/appcenter-reviews/runs/6137393588?check_suite_focus=true#step:6:20)
          buildsystem: meson
```

It looks like you're already using FileChooserNative [here](https://github.com/alainm23/planner/blob/6afe275f2245efa39a8a936fa7b9e6527e8a6114/src/Services/MigrateV2.vala#L521) which uses the file chooser portal. The portal runs out of process, so Planner doesn't need any explicit filesystem permission when opening files like this. It would only need permissions if you're trying to access files directly without using the filechooser

Since Planner uses an online API for Todoist, it should at least have `social-info` set to `mild` as defined by [OARS generate](https://hughsie.github.io/oars/generate.html):
```
  <content_rating type="oars-1.1">
    <content_attribute id="social-info">mild</content_attribute>
  </content_rating>
```

I noticed you have gresourced a bunch of icons that are `-light` and `-dark`. If you instead suffix one of these versions as `-symbolic`, Gtk will recolor them for you. For colored icons, you can either manually modify the svg to include "error", "warning", "success" classes, or with Gtk.CSS you can add `accent` and a color name like "red", "blue", "green", etc and you'll get colors that fit in the palette and adjust brightness for contrast automatically

PR #321, opened 2022-02-10
com.github.alexkdeveloper.pasgen 1.0.0

alexkdeveloper:
## Review Checklist
- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,8 +36,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: parse
 
+    strategy:
+       matrix:
+         arch: [x86_64, aarch64]
+       # Don't fail the whole workflow if one architecture fails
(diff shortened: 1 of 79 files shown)


danrabbit (approved):
Looks good! Thanks for your patience with the review time.

A couple of non-blocking comments for your next update:

You don't have to include OARS tags for anything that has a value of `none`. So since your app doesn't have any content warnings, you can simply self-close the OARS parent tag like `<content_rating type="oars-1.1" />`

I would recommend using the `suggested-action` style class instead of custom CSS for your "Generate Password" button. The label does currently pass WCAG AA, but using the built-in style class would let users choose their own accent color and bring it inline with the switch accent color

PR #328, opened 2022-03-06
Update Pebbles to 2.0.5

SubhadeepJasu:
Repository [SubhadeepJasu/Pebbles](https://github.com/SubhadeepJasu/Pebbles)

## Review Checklist

- [x] App Opens
- [x] Does what it says
- [x] Categories match

## AppData

- [x] Matches Description
- [x] Matches Screenshot
- [x] OARS info matches

## Flatpak

- [x] Uses elementary runtime

The system talk string still needs to be removed. I'll have to test that for the next release.

### DIFF
--- a/applications/com.github.subhadeepjasu.pebbles.json
+++ b/applications/com.github.subhadeepjasu.pebbles.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/SubhadeepJasu/pebbles.git",
-  "commit": "fcc6a92f7676fd1e083ce93fc653b1c25f13a3ad",
-  "version": "2.0.4"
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
>The system talk string still needs to be removed. I'll have to test that for the next release.

Since you bumped the runtime version, it should work exactly the same as before without the arg. That said, I don't believe that's a blocking issue. Just remember to remove it soon. :smile:

cassidyjames:
@SubhadeepJasu oh another note as long as we're here: you don't need to include any OARS tags which have a value of `none`; as of OARS 1.1, missing values are implied to be `none`. Cleaning those up can help the non-none values stand up which can also help us review that section a bit faster. :wink:

SubhadeepJasu:
Sure. My bad 😅.

PR #330, opened 2022-03-08
Update com.github.tenderowl.norka to 1.0.1

amka:
https://github.com/TenderOwl/Norka/compare/1.0.0...1.0.1

Fixes and translations.

## Review Checklist
- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime


cassidyjames (approved):
As long as we're here: you don't need to include any OARS tags which have a value of `none`; as of OARS 1.1, missing values are implied to be `none`. Cleaning those up can help the non-none values stand out which can also help us review that section a bit faster. :wink:

PR #332, opened 2022-03-13
Create com.github.phoneybadger.trimmer

adithyankv:
## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.phoneybadger.trimmer.json
+++ b/applications/com.github.phoneybadger.trimmer.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/phoneybadger/trimmer.git",
+  "commit": "886edaa09cda8f78cb703cb7d0e94f2eb315b1d4",
+  "version": "1.0.0"
+}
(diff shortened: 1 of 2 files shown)


danrabbit:
Hey @adithyankv  thanks for your patience with the longer than usual review time here! It looks like the release tag `1.0.0` in GitHub is pointing to a different commit than you've listed here. You'll probably want to either redo that tag or release a `1.0.1` so those match up

Other than that everything looks good! I wasn't able to successfully trim a video but that might be a quirk of my development setup. It might be worth grabbing the flatpak bundle from the build here just to confirm on your end that it is indeed working as expected

:tada:

adithyankv:
Hey @danrabbit , thanks for the review.
I've created a new `1.0.1` release so that it matches with the commit. And grabbing the flatpak build and installing on my side does seem to work as expected.

danrabbit (approved):
Perfect, thank you! Let's get this published :rocket:

PR #343, opened 2022-04-13
Update com.github.phoneybadger.trimmer to 1.0.2

adithyankv:
Fixes trimming bug

### DIFF
--- a/applications/com.github.phoneybadger.trimmer.json
+++ b/applications/com.github.phoneybadger.trimmer.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/phoneybadger/trimmer.git",
-  "commit": "af85cb9b47157b024bc06540cf80f3fbf45d9d2a",
-  "version": "1.0.1"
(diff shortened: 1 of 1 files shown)


danrabbit:
Hey @adithyankv unfortunately we can't redistribute FFMPEG due to patent restrictions in the US. You'll have to keep that as an extension to be able to distribute your app through AppCenter

adithyankv:
> Hey @adithyankv unfortunately we can't redistribute FFMPEG due to patent restrictions in the US. You'll have to keep that as an extension to be able to distribute your app through AppCenter

Oh okay. 

Then this is going  to be a tricky one. Trimmer does require the extension `org.freedesktop.Platform.ffmpeg-full` to run, but then I cannot seem to find a way to auto-download it when  the app is installed. I believe it can only be auto-downloaded if the extension is in the same repository as the application, which in this case it isn't. I guess I can make the user manually install the extension, but then that would be horrible UX. 

Let me see if I can find some other elegant way to do this.

Thanks for the review

PR #351, opened 2022-05-12
Add com.github.oowoosh0.pomodoro

Oowoosh0:


danrabbit (changes requested):
Hey thanks for your submission! I have just a couple of things to note here:

You're missing a release tag in your app data. We need a release tag that matches the version number you're trying to publish as described here: https://docs.elementary.io/develop/appcenter/publishing-updates#release-descriptions

Your .desktop file only has the category "utility", but your app probably should have the "office" category

This isn't a requirement, but I would recommending using version `7` of the platform instead of `6.1` so that your app will be ready when OS 7 is released. It will still be able to be published for 6.1 users when using the 7 platform

Otherwise looks good, nice work :tada:

Oowoosh0:
Thanks for your quick review, I added the release info to the app data and changed the category in the desktop file to office.
also updated the flatpak runtime to 7.

danrabbit (approved):
Looks good! Thank you for the quick turnaround as well :rocket:

PR #356, opened 2022-05-18
Update com.github.rajsolai.textsnatcher v2.0.0

RajSolai:


danrabbit (approved):
Way to tighten up those permissions!

Just a heads up, a new version of the Flatpak platform is out so you'll want to update from `6.1` to `7` before your next release :)

PR #361, opened 2022-05-28
com.github.brain-child.moobo 0.1.3

pierrefabarius:


danrabbit (approved):
Looks good!

Just so you know, a new runtime version has been released, so you probably want to update from version `6.1` to `7` before your next release.

I would also recommend switching from grabbing raw key events to using GLib.Actions. This will make it quite a bit more straightforward to support shortcuts for different types of keyboard layouts. Check out the docs here: https://docs.elementary.io/develop/apis/actions#glib.simpleaction

PR #371, opened 2022-06-12
Time Limit 3.0.0

marbetschar:
- Migrated to Gtk4
- Update to elementary Platform v7


danrabbit (approved):
Nice job porting to Gtk4!

I'm not going to hold this update back, but I think instead of adding a sandbox hole for loginmanager, that you can use libportal to monitor for session state changes: https://valadoc.org/libportal/Xdp.Portal.session_state_changed.html

danrabbit:
Filed an issue for lack of documentation here: https://github.com/elementary/docs/issues/176

It looks like there's also https://valadoc.org/gtk4/Gtk.Application.screensaver_active.html

PR #373, opened 2022-06-18
com.github.ryonakano.reco 3.3.1

ryonakano:
Changes summary: https://github.com/ryonakano/reco/releases/tag/3.3.1


ryonakano:
Ah I thought I'd already lessened the scope but that seems to be a fix in another project. :sweat_smile: Will fix in the next version, thank you for your info!

PR #380, opened 2022-06-20
com.github.jeremypw.gnonograms version 2.1.1

jeremypw:


danrabbit (changes requested):
Oops, it looks like you've got a copy/paste error in your appdata:
` <launchable type="desktop-id">io.elementary.music.desktop</launchable>`

jeremypw:
Huh! I always end up producing about 3 releases before I get one right!  Thanks for spotting that.

PR #386, opened 2022-06-27
com.github.sadhvikbathini.cson.json

sadhvikbathini:
## Review Checklist
- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/com.github.sadhvikbathini.cson.json
+++ b/applications/com.github.sadhvikbathini.cson.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/sadhvikbathini/cson.git",
-  "commit": "b2a5863b472bfa3a1d6da55540802bc3ee7d206c",
+  "commit": "dfcfa03ba821d178593ad287404f0e8284f4f5d2",
   "version": "0.0.1"
 }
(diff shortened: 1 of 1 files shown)


danrabbit (changes requested):
Hey thanks for submitting your app for review. It looks like there are a few problems to resolve before we can publish your app.

The screenshot URLs in your appdata don't seem to be valid. For example: https://raw.githubusercontent.com/sadhvikbathini/cson/data/assets/screenshots/cson.png

The screenshots [here (for example)](https://github.com/sadhvikbathini/cson/blob/main/data/screenshots/cson-screen.png) seem to depict an older version of elementary OS and not what your app looks like running from the Flatpak on the currently supported release of elementary OS.

Please keep in mind that screenshots should depict only your app, not other apps as seen [here](https://github.com/sadhvikbathini/cson/blob/main/data/screenshots/terminal.png).

Your appdata doesn't contain a `releases` tag as described here: https://docs.elementary.io/develop/appcenter/publishing-updates#release-descriptions

It looks like your app is maybe relying on installing assets to the users' home directory and hardcoding paths to those assets. Your app should instead use GResources for custom assets. See: https://docs.elementary.io/develop/apis/gresource

It looks like your app accesses the command line to save and retrieve state. This should be done using GSettings instead. See: https://docs.elementary.io/develop/apis/state-saving and https://valadoc.org/gio-2.0/GLib.Settings.html

Your app is also using an old version of the elementary Flatpak runtime. Please update from version `6` to version `7`

danrabbit:
Closing since the submitter never responded to requested changes

PR #393, opened 2022-07-16
Update Comic Sticks to version 1.6.4.

rkoesters:
Release notes: https://github.com/rkoesters/xkcd-gtk/releases/tag/1.6.4


zeebok (approved):
Things look good here. Only thing I see is that there is no `launchable` tag in the AppData. It is optional but elementary recommends having it and you can always add it in the next release if you decide to.

PR #394, opened 2022-07-18
com.github.elfenware.obliviate v2.4.0

dar5hak:


zeebok:
Looking through it seems all that is missing is a `launchable` tag in your appdata. I do notice a `provides` and `binary` tag used though. @danrabbit  is that an allowed alternative or does he need `lauchable` specifically?

danrabbit:
I believe that the `launchable` tag might be required to show the "Open" button in AppCenter, but I could be wrong about that. It's optional but recommend :)

dar5hak:
Okay, I tested it. The "Open" button in AppCenter shows up and works with the current published version.

That said, I will add the tag if it's recommended. Do you need me to add it to this version, or is it okay if I release it later?

zeebok:
I think if it is working alright in appcenter you can add in the next release

zeebok (approved):

dar5hak:
Thank you, Ryan!

PR #398, opened 2022-08-13
com.github.juarezfranco.tarefas-desktop 1.1.1

juarezfranco:
## Review Checklist
- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [x] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.juarezfranco.tarefas-desktop.json
+++ b/applications/com.github.juarezfranco.tarefas-desktop.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/juarezfranco/tarefas-desktop.git",
+  "commit": "3983b01347d4349b420be9c18775b257970a389f",
+  "version": "1.1.1"
+}


danrabbit (changes requested):
Hey thanks for your submission! There's a few things that need changing before we can publish your app in AppCenter:

- [ ] Your app's name is the same as a default app, which is potentially confusing. Your app's name should be a unique brand name and not a generic name.
- [ ] Your screenshot looks like it was taken on an older version of elementary OS. You'll want to update your screenshot to show your app running on the latest stable release, which is currently 6.1
- [ ] Your appdata is missing a release tag for version 1.1.1

Some other recommendations, but not required:
- [ ] Your app is currently targeting elementary flatpak platform 6.1 which is okay, but 7.0 is available so I would recommend updating to that
- [ ] Your custom colors meet WCAG A level contrast which is the bare minimum, but I would recommend trying to meet WCAG AA so that your app listing will be legible to the most amount of people
- [ ] I would recommend using the `Office` category instead of `Utility` in your .desktop file so that it will be easier for people to discover your app in AppCenter
- [ ] Instead of a manual light and dark switch, I would recommend making use of system dark style preference, which is documented here: https://docs.elementary.io/develop/apis/color-scheme#setting-a-dark-style

Thanks again and let me know if you have any questions!

ryonakano:
Closing due to inactivity over 3 years after the above changes request. Feel free to reopen if you address them.

PR #399, opened 2022-08-16
com.github.ryonakano.atlas 1.0.1

ryonakano:
Just a couple of updates like translations and the dependency.

I know now it is recommended to use the version 7 of elementary Flatpak runtime, but if we try to perform that change, I also need to migrate to GTK 4 and libshumate at the same time, which is challenging for me. This is because the GNOME Flatpak runtime 42, which elementary Flatpak runtime depends on, removes the necessary dependencies. So I publish a release with 6.1 version of the runtime meanwhile.

See https://github.com/ryonakano/atlas/issues/26 for more details.

### DIFF
--- a/applications/com.github.ryonakano.atlas.json
+++ b/applications/com.github.ryonakano.atlas.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ryonakano/atlas.git",
-  "commit": "1fe825fc6ddfb66e8e1da22c777f82de0e8c9958",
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Hey that's fine you're welcome to keep building against the old runtime until the new shumate deps etc are available and you can port. You also have the option of bundling clutter in your manifest which is what Videos does for example: https://github.com/elementary/videos/blob/master/io.elementary.videos.yml#L28

PR #401, opened 2022-09-07
Update com.github.alexkdeveloper.dfc to 1.0.4

alexkdeveloper:


danrabbit (approved):
Looks good!

One small suggestion I have for you: in the next version of AppCenter, we'll start exposing screenshot captions like this:

Screenshot from 2022-09-06 14 44 14


So I would suggest adding captions to your screenshots since it's a great way to explain the various features of your app and making your screenshot more accessible for low vision users. You can read more here: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

PR #403, opened 2022-09-22
Update com.github.fleury08.prettifier 1.3.0

fleury08:


danrabbit (approved):
Nice work, tight sandbox!

A couple small suggestions for AppData in your next release: The next version of AppCenter will show captions for screenshots on your app info page, so you might want to consider adding those! Also, we're migrating from the custom appcenter color keys to the upstream branding key that was introduced in AppStream. You can read more about both of those things here: https://docs.elementary.io/develop/writing-apps/our-first-app/metadata

fleury08:
Thanks ☺️ I will look at the suggestions you mentioned and implement them on the next release. 😏 Dne 22. 9. 2022 23:56 napsal uživatel Danielle Foré ***@***.***>:
@danrabbit approved this pull request.

Nice work, tight sandbox!
A couple small suggestions for AppData in your next release: The next version of AppCenter will show captions for screenshots on your app info page, so you might want to consider adding those! Also, we're migrating from the custom appcenter color keys to the upstream branding key that was introduced in AppStream. You can read more about both of those things here: https://docs.elementary.io/develop/writing-apps/our-first-app/metadata

PR #408, opened 2022-09-28
Update com.github.alexkdeveloper.tictactoy to 1.0.5

alexkdeveloper:


alexkdeveloper:
There is only one screenshot in the description of the application, so I did not add a signature to it. Do I understand correctly that captions to screenshots should be added if there are several of them?

danirabbit (approved):
Looks good!

danirabbit:
@alexkdeveloper I would recommend adding a caption regardless since the label also helps make your app listing more accessible to low vision users

PR #411, opened 2022-10-10
com.github.ryonakano.reco 4.0.0

ryonakano:
- Diff from the previous release: https://github.com/ryonakano/reco/compare/3.3.1...4.0.0
- Release note: https://github.com/ryonakano/reco/releases/tag/4.0.0

### DIFF
--- a/applications/com.github.ryonakano.reco.json
+++ b/applications/com.github.ryonakano.reco.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ryonakano/reco.git",
-  "commit": "fdd3e6a2ec7da0bb3636a7b779f6fd54ffbbc035",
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Looks good! Nice job porting to Gtk4 and tightening that sandbox hole!

Couple suggestions for future versions:

Future version of AppCenter will use the upstream `<branding>` tag, so I would recommend adding that now. More info here: https://docs.elementary.io/develop/writing-apps/our-first-app/metadata#branding

Also, instead of installing your custom icon via meson, I would recommend using GResource to include it. More information here: https://docs.elementary.io/develop/apis/gresource#icons

ryonakano:
Wow nice, thank you for your info! :tada:

PR #412, opened 2022-10-16
add: Gradience (com.github.GradienceTeam.Gradience) - v0.3.1

0xMRTT:


danirabbit:
Hey thanks for your submission! It's my understanding that this tool is specifically for Adwaita right? Since elementary OS doesn't use Adwaita, this seems like an ill fit for AppCenter since it would have no effect for elementary OS users

0xMRTT:
That can customize more things than just Adwaita but yes you are maybe right

danirabbit:
As far as I can tell, it looks like it only effects GNOME, is that right? For example we don't use GNOME Shell or GDM etc.

It also looks like this flatpak manifest builds against the GNOME runtime, but we require everything in the appcenter repo to build against the elementary runtime, use the elementary stylesheet, show elementary defaults in screenshots, etc

0xMRTT:
You are right.

danirabbit:
Thanks for submitting anyways! ❤️

PR #417, opened 2022-11-09
Update com.github.alexkdeveloper.dfc to 1.0.6

alexkdeveloper:


danirabbit (approved):
Hey thanks for your patience! There were some errors in our publishing pipeline but it should all be resolved now. Nice work updating to FIleChooserNative and tightening your sandbox permissions!

One thing to note for you next release is that we have a new platform available `7.1`!

PR #425, opened 2022-12-09
Add Gradebook

leolost2605:
A little application to keep track of your grades.
My first app so if i missed anything or disregarded best practices i would love some feedback :)

## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/io.github.leolost2605.gradebook.json
+++ b/applications/io.github.leolost2605.gradebook.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/leolost2605/Gradebook.git",
-  "commit": "dae52b1",
-  "version": "v1.0.0"
+  "commit": "ef75dbcb5cdae089be4e80d846dc52a12f775d01",
+  "version": "v1.0.1"
 }
--- a/applications/io.github.leolost2605.gradebook.json
+++ b/applications/io.github.leolost2605.gradebook.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/leolost2605/Gradebook.git",
+  "commit": "dae52b1",
+  "version": "v1.0.0"
+}
--- a/applications/io.github.leolost2605.gradebook.json
+++ b/applications/io.github.leolost2605.gradebook.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/leolost2605/Gradebook.git",
-  "commit": "ef75dbcb5cdae089be4e80d846dc52a12f775d01",
-  "version": "v1.0.1"
+  "commit": "1fbefa95eb502e44567d82b6db146e1fb9c0e3e8",
+  "version": "v1.0.2"
 }


danirabbit (changes requested):
Hey thanks for your submission!

It looks like your commits aren't lining up here. You have a tag which is commit `b4be49feb7fed9329fd76788a00d706ea369ab55`, but the manifest in that commit is pointing to `dae52b1`. It would probably be better to keep things like you had them in [this commit](https://github.com/leolost2605/Gradebook/commit/27af1bb3831dd59954f8427e038d9e310a418543) so that you're building the same commit as the tag

It looks like your app is also using the GNOME Platform, but apps submitted in AppCenter must use the elementary Platform. This is what's causing the current build failure since the GNOME platform isn't available in the AppCenter remote

In your desktop file, I would recommend using the category `Education`. The `Utility` category ends up as kind of a junk drawer and it will make it harder for people to find your app. We don't do anything with the `GNOME` category in AppCenter and I would probably remove this in case folks use the AppCenter remote and their client does use this to filter out Adwaita-style apps. You can find a list and explainer of categories [here](https://specifications.freedesktop.org/menu-spec/latest/apa.html) as well as additional subcategories [here](https://specifications.freedesktop.org/menu-spec/latest/apas02.html)

It doesn't look like the URLs for your screenshots are valid. It seems like they would point to screenshots in your toplevel directory but there's nothing there. I would also recommend versioning these URLs so that if you update the screenshot in the future it won't change it out in AppCenter before that version is published. I would also recommend adding `caption` tags to your screenshots to make your page more accessible for folks with low vision and to help directly explain its features. Check out this section of our developer guide: https://docs.elementary.io/develop/writing-apps/our-first-app/metadata#screenshots

Let me know if you have any questions!

leolost2605:
Thank you so much for the detailed Feedback! I will work on those fixes right away

danirabbit (changes requested):
In your desktop file, [this](https://github.com/leolost2605/Gradebook/blob/ef75dbcb5cdae089be4e80d846dc52a12f775d01/data/io.github.leolost2605.gradebook.desktop#L9) should be an icon name, not a file name, so you don't want to include the extension `.svg` here

Your screenshots seem to show an Adwaita-style app with GNOME default settings for window controls for example. Your screenshots need to use elementary default settings and your app needs to use the elementary stylesheet as per https://docs.elementary.io/develop/appcenter/publishing-requirements#metainfo-and-oars

Don't forget to add a new release tag to your metainfo file match the version that you're submitting :)

leolost2605:
Oh i'm sorry i didn't know adwaita was forbidden entirely. As i am relying pretty heavily on adwaita i'm gonna close the request for now and see what i can do. Thanks again for all the feedback you have given me!

danirabbit:
Hey @leolost2605 you're more than welcome to use LibAdwaita in your app, but it looks like subclassing Adw.Application instead of Gtk.Application is forcing your app to use the Adwaita stylesheet. From what I can see your app does compile using Gtk.Application and the biggest style issue there is manually adding the css class "titlebar" to your headerbar since you're not using `Gtk.Window.set_titlebar`

leolost2605:
Hi, thanks for the info i'll look into it and reopen the PR when i get it fixed.

PR #430, opened 2022-12-29
Update com.github.childishgiant.mixer to 1.1.0

ChildishGiant:
This release adds 2 translations and fixes a flatpak permissions issue on some distros.  
https://github.com/ChildishGiant/mixer/releases/tag/1.1.0


danirabbit (approved):
Hopefully we'll be able to scope down the device permission there in the future with maybe Pipewire or something in LibPortal

Something I noticed here is that you're still using the 6.1 runtime, there is actually a 7.1 runtime available now so make sure to update that on your next release to get the latest library fixes etc :)

PR #431, opened 2022-12-29
Update Actions

danirabbit:
* Build using the 7.1 runtime container
* Use flatpak-builder v5
* Checkout v3


meisenzahl (approved):
Let's go 🚀

PR #450, opened 2023-04-24
Update com.github.phase1geo.minder.json

phase1geo:
Creating 1.15.0 release

### DIFF
--- a/applications/com.github.phase1geo.minder.json
+++ b/applications/com.github.phase1geo.minder.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/phase1geo/Minder.git",
-  "commit": "3301377ff575ff435f29846bfdd9f87ab5b8d8fc",
-  "version": "1.14.0"
+  "commit": "21d9622596120a7860a82c985b8849828ec31713",
+  "version": "1.15.0"
 }


phase1geo:
In an attempt to pre-answer a question that I might get about enabling "filesystem=home" in the Flatpak manifest.  This was user requested and is necessary because the application retains tab state between invocations where each tab contains the contents of a file that may be located in the user's home directory.  On application start, previously opened files are read by the application.  The filesystem portal is used to open new files after the application is opened and is used to save files.

If there is a better way to load user files into the application without requiring the "filesystem=home" value to be used, I am open to suggestions.  But using this mechanism seems to be the most straightforward approach to solving this problem, and users don't seem interested in enabling this on their system when the application requires it and not having it enabled means unexpected results in the application.

danirabbit (approved):
Huge release, great job!

About the home permission, I believe file chooser portal permissions should persist between sessions, but I'm not totally sure. Expanding the sandbox isn't great, we'll be showing more warnings in AppCenter soon about sandbox holes so keep that in mind in the future there will be a warning on your AppInfo page, but other than that it's up to you and your users how you want to manage trust :)

PR #455, opened 2023-05-06
Update com.github.phase1geo.annotator.json

phase1geo:
Updating to 1.2.0 release.

### DIFF
--- a/applications/com.github.phase1geo.annotator.json
+++ b/applications/com.github.phase1geo.annotator.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/phase1geo/Annotator.git",
-  "commit": "e4bc474b201e8ebccc7f91a68596bf692f79e6f1",
-  "version": "1.1.1"
+  "commit": "356cfe5c63917d0f67398d73ea16b0ced4bda278",
+  "version": "1.2.0"
 }


danirabbit (approved):
Hey nice update! A few notes here:

We now support the `issues` tag in appdata, so I highly recommend making use of that tag to describe issue fixes instead of including them in the `description` tag in your release notes. More info on that here: https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Releases.html#tag-release-issues

`home` is a really broad filesystem access permission. Generally filesystem permissions shouldn't be needed if you're using the file chooser portal. Be aware that soon in AppCenter we'll be warning about sandbox holes like this. I would really recommend tightening that permission as much as possible before your next release.

Similarly with the screenshot permission, your app shouldn't directly access the screenshot interface. You should instead use the screenshot portal: https://valadoc.org/libportal/Xdp.Portal.take_screenshot.html

And again we'll be showing a warning for apps that don't use the portal in the future, something like "This app can take screenshots without your permission". It's a really serious sandbox hole, so make sure you switch over to using the portal.

phase1geo:
Thank you for the suggestions. I will look into those shortly.

phase1geo:
@danirabbit Out of curiosity, I wasn't able to get the screenshot portal to work properly on elementary OS.  The only way that I could get screenshots to work was to install the xdg-desktop-portals-gnome package.  Unfortunately, this shows the Gnome screenshot utility instead of the elementary UI.  Is there a way to get this to work on elementary?

danirabbit:
Ah apologies, you're right it looks like we still have an open issue report here: https://github.com/elementary/screenshot/issues/140

So ignore that for now :sweat_smile:

PR #458, opened 2023-05-23
Update com.github.phase1geo.minder.json to 1.15.2

phase1geo:


danirabbit (approved):
Looks good, just want to let you know that we have `issues` tag support in AppCenter, so if you use that in your release notes you'll get a nicely formatted issues list with links etc. Check out: https://docs.elementary.io/develop/appcenter/publishing-updates#issues

PR #461, opened 2023-06-04
Create com.github.dcharles525.Tempo.json

dcharles525:
## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.dcharles525.Tempo.json
+++ b/applications/com.github.dcharles525.Tempo.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/dcharles525/Tempo.git",
   "commit": "86b81689ac8109145485aa9fe8ef35edaa9834b1",
-  "version": "1.0.0"
+  "version": "1.1.0"
 }
--- a/applications/com.github.dcharles525.Tempo.json
+++ b/applications/com.github.dcharles525.Tempo.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/dcharles525/Tempo.git",
+  "commit": "86b81689ac8109145485aa9fe8ef35edaa9834b1",
+  "version": "1.0.0"
+}


danirabbit (changes requested):
Hey thanks for your submission!

I have some requested changes below, but most importantly your app doesn't seem to open or run. I think this is probably due to the mismatch in capitalization with your app_id [here](https://github.com/dcharles525/Tempo/blob/86b81689ac8109145485aa9fe8ef35edaa9834b1/src/Tempo.vala#LL8C24-L8C52). It also seems like your app is trying to access filesystem locations that it doesn't have permission for [here](https://github.com/dcharles525/Tempo/blob/86b81689ac8109145485aa9fe8ef35edaa9834b1/src/Libraries/Sense.vala#L32). Keep in mind that access to `/sys` is blacklisted as documented [here](https://docs.flatpak.org/en/latest/sandbox-permissions.html)

Make sure you're building your app with flatpak builder and testing it running from the flatpak before submitting :)

In your flatpak manifest, you're using platform 6.1. This is a very old platform version, so your app would immediately show in AppCenter as being outdated. The current version of the platform is 7.2, so please make sure to update that.

The screenshot URL in your metainfo file is invalid, so please update that to a valid URL :) The screenshot that I see in your git repo also seems to include things other than your app. Your screenshot should [only contain your app](https://github.com/micahilbery/elementary-icon-templates/tree/master/icon-templates/Square%20Icon). The easiest way to get a window only screenshot is to secondary click on your app's title bar and select "Take screenshot".

You're also missing some [required tags](https://specifications.freedesktop.org/menu-spec/latest/apas02.html) in your metainfo file. Those are:
- [ ] developer_name
- [ ] [content_rating](https://hughsie.github.io/oars/generate.html)

Although not strictly required, I would highly recommend you include [`url`](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url) tags so that folks know where to report issues and get help

More appropriate categories for your .desktop file would probably be "Monitor" and "System". See the categories specification here: https://specifications.freedesktop.org/menu-spec/latest/apas02.html

I would also recommend looking at basing your app icons on Micah's great [icon templates](https://github.com/micahilbery/elementary-icon-templates/tree/master/icon-templates/Square%20Icon) so that they appear the right size and shape and with the right lighting and shadows to fit in with other apps.

In your Meson file, it looks like you've duplicated installing your desktop and metainfo files. `i18n.merge_file` already includes an install step, so you don't also need to run `install_data` first.

For custom CSS styles, you should include those using GResource. At the moment we only have documented resourcing icons, but there's relevant document [here](https://docs.elementary.io/develop/apis/gresource) for creating a gresource file. And you can use that with [`Gtk.CssProvider.load_from_resource`](https://valadoc.org/gtk+-3.0/Gtk.CssProvider.load_from_resource.html)

dcharles525:
@danirabbit thanks for all the help with this. I think I have resolved everything now!

danirabbit (changes requested):
Thanks for resolving those! It looks like we're still having some issues here:

 - [ ]  Your screenshot URLs in Metainfo are still invalid. They don't point to your github repo: https://github.com/dcharles525/Tempo/blob/main/data/Tempo.metainfo.xml.in#L23C44-L23C44
 - [ ] There should only be one `<screenshots>` tag that contains all `<screenshot>` tags
 - [ ] It is only valid to have one `<screenshot>` tag with `type="default"`. This is the default screenshot of your app, so there can't be more than one default :)
 - [ ] Your application ID is still mismatched. capitalization should match the ID in your Flatpak manifest, desktop file, etc: https://github.com/dcharles525/Tempo/blob/6b4a6f57514cf800d2047ca34efe34c0cd6498da/src/Tempo.vala#L6C23-L6C52
 - [ ] You really should not have this session bus line [here](https://github.com/dcharles525/Tempo/blob/6b4a6f57514cf800d2047ca34efe34c0cd6498da/com.github.dcharles525.Tempo.yml#L23). Your app is failing to register on startup because the app ID doens't match because of the incorrect capitalization
 - [ ] I would still recommend including the "Monitor" category in your .desktop file in addition to "System"
 - [ ] And again I want to link the app icon templates [here](https://github.com/micahilbery/elementary-icon-templates/tree/master/icon-templates/Square%20Icon) since your icons are incorrectly sized and do not match the style of other platform icons
 - [ ] You also don't have any `release` tags in metainfo which are required: https://docs.elementary.io/develop/writing-apps/our-first-app/metadata#releases

PR #489, opened 2023-10-22
Update com.github.ztefn.haguichi to 1.146.1

ztefn:
Rebased [elementary](https://github.com/ztefn/haguichi/tree/elementary) branch on top of master, from commit [6e6f411 to d46f457](https://github.com/ztefn/haguichi/compare/6e6f411...d46f457). Runtime updated to 7.2.

Release tag: [1.146.1](https://github.com/ztefn/haguichi/releases/tag/1.146.1)

### DIFF
--- a/applications/com.github.ztefn.haguichi.json
+++ b/applications/com.github.ztefn.haguichi.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ztefn/haguichi.git",
-  "commit": "34822b7a3324bf6934d54722c59040133d375806",
-  "version": "1.146.0"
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Looks really good, nice work! Thanks for adding screenshot captions!

One thing I'd recommend to investigate using portals more.
- There is a built-in method to Gtk.Application for monitoring session state: https://valadoc.org/gtk4/Gtk.Application.screensaver_active.html
- I would also recommend looking at using the background portal to make sure we're properly asking for permission to run in the background and providing a reason/status etc: https://valadoc.org/libportal/Xdp.Portal.request_background.html

PR #501, opened 2024-02-06
Whaler 1.2.2

sdv43:
## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime


danirabbit (approved):
:tada:

PR #505, opened 2024-02-20
Create io.github.xchan14.larawan.json

xchan14:
## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/io.github.xchan14.larawan.json
+++ b/applications/io.github.xchan14.larawan.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/xchan14/larawan.git",
-  "commit": "063f5cf14cd32cc68400c320b608ae44c808f5e6",
+  "commit": "92120ec8342475b1ad34be131565f36dc946f3bf",
   "version": "1.0.0"
 }
--- a/applications/io.github.xchan14.larawan.json
+++ b/applications/io.github.xchan14.larawan.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/xchan14/larawan.git",
-  "commit": "92120ec8342475b1ad34be131565f36dc946f3bf",
+  "commit": "121cc9df9af114b60c0a1c2739280de970862d45",
   "version": "1.0.0"
 }
--- a/applications/io.github.xchan14.larawan.json
+++ b/applications/io.github.xchan14.larawan.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/xchan14/larawan.git",
+  "commit": "063f5cf14cd32cc68400c320b608ae44c808f5e6",
+  "version": "1.0.0"
+}


danirabbit (changes requested):
Hey thanks for your submission! I haven't done a super thorough review yet, but on first glance there's a few things that need fixing before we can publish:

- [x] Your metainfo file contains placeholder text instead of a description
- [ ] Screenshots show your entire desktop, they should only show your app
- [ ] You app's description in metainfo and in your desktop file should not mention elementary or elementary OS
- [ ] Your metainfo doesn't include any developer information such as your name and where folks can report issues or get help
- [ ] I would highly encourage you not to add home folder permissions to your flatpak manifest and instead use the FileChooser portal to access files. Your app will be badged with a warning that it accesses the home folder if you add permissions in this way
- [ ] There is a 7.3 runtime available, I would recommend updating to that :)

See https://docs.elementary.io/develop/appcenter/publishing-requirements for some more specific guidance

Also just a little tip, I can't remember if it's the 7.2 or 7.3 runtime, but Granite will now automatically load a gresourced stylesheet named "Application.css" (capitalization matters) when you run Granite.init, so there's no need to do that manually as in https://github.com/xchan14/larawan/blob/38cdb432c349c6309fff104b8074b91a219316ea/src/Application.vala#L31

danirabbit:
Hey @xchan14 thanks for the quick turnaround! Everything is looking really good except it looks like there is no release tag in your metainfo file. Apologies for not catching that in the initial review. See the docs here for more information about creating a release tag: https://docs.elementary.io/develop/writing-apps/our-first-app/metadata#releases

Otherwise this seems okay to publish once we have that release tag! Congrats :tada: 

A couple things I noticed here that aren't requirements but just advice:

You should init Granite and apply styles in startup, [not in activate](https://github.com/xchan14/larawan/blob/92120ec8342475b1ad34be131565f36dc946f3bf/src/Application.vala#L23). Activate runs every time your app is launched from the dock or applications menu, including when folks click on the icon to switch windows. You really only want to do this once during your app's lifecycle, which is when the startup function runs.

There are built in icons that you can use with your app [instead of using an emoji](https://github.com/xchan14/larawan/blob/92120ec8342475b1ad34be131565f36dc946f3bf/src/views/MainWindow.vala#L77C9-L77C57). There's an app in AppCenter called [IconBrowser](https://appcenter.elementary.io/io.elementary.iconbrowser/) that you can use to view all the available system icons and get a code snippet. In this case I would recommend using `Gtk.Button.from_icon_name ("open-menu")`. Don't forget you should add a tooltip to image buttons so that there's something for the screen reader to read out to describe them

xchan14:
Hi @danirabbit, done adding the release tag. As well as the your advice. Thank you!

PR #519, opened 2024-07-18
Update Detective to 1.1.0

leolost2605:
Release notes: https://github.com/leolost2605/detective/releases/tag/1.1.0

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/io.github.leolost2605.detective.json
+++ b/applications/io.github.leolost2605.detective.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/leolost2605/detective.git",
-  "commit": "2547ff344a9fb65f6afddf5601beed6e8cf2ebfd",
-  "version": "1.0.2"
+  "commit": "635fdd7b3082382fdfb749c0cf7b1e715817640f",
+  "version": "1.1.0"
 }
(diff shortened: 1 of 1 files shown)


danirabbit (changes requested):
I think we probably explicitly don't want third party apps to be able to access Gala's desktop integration features as a matter of security. Especially not without like explicit consent. Apps not being able to focus themselves or see other running apps or windows on Wayland is a feature

leolost2605:
I kinda disagree but I updated it :)

danirabbit:
It looks like it's still touching gala here? https://github.com/leolost2605/detective/compare/1.0.2...08a375610aad69458cfd0e085f517fa6212c41af

Accessing this interface without explicit user consent is a pretty serious privacy issue imo. If apps want to be able to read this kind of information, it needs to be through a portion where users can say no

leolost2605:
I'm not sure whether I missed something but the main part (i.e. DesktopIntegration.vala and the stuff that uses it) can (and is in the AppCenter version) be disabled via a meson build option, so that it's not even compiled in.

danirabbit:
Ah sorry I didn't catch that there was a meson flag there. Will re-review Monday!

danirabbit (approved):

PR #523, opened 2024-08-12
com.github.ryonakano.pinit 2.1.1

ryonakano:
Publishing to AppCenter again thanks to `Granite.init ()`, which can force elementary stylesheet instead of Adwaita one.

- Diff from the previous release: https://github.com/ryonakano/pinit/compare/1.4.1..2.1.1
- Release note: https://github.com/ryonakano/pinit/releases/tag/2.1.1

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime


danirabbit (approved):
:tada:

PR #532, opened 2024-08-29
Create io.github.danirabbit.harvey.json

danirabbit:
## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime


zeebok (approved):
This all looks good. Like the redesign too! I did notice that the ratio label is highlighted when you launch (at least on OS7 where I tested) and I'm not sure if that is intentional/expected or not?

danirabbit:
@zeebok hm I don't think I'm actually doing anything about manually setting focus but I suppose it should be in one of the entries instead maybe? Can you open an issue report for me? 🙏🏼

PR #534, opened 2024-10-08
Add App Generator

alainm23:
## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/io.github.ecommunity.app-generator.json
+++ b/applications/io.github.ecommunity.app-generator.json
@@ -1,0 +1,6 @@
+{
+    "source": "https://github.com/elementary-community/app-generator.git",
+    "commit": "e3496d905dc54ba1eab9a67e9eddf9f6a6907345",
+    "version": "1.0.0"
+  }
+  


danirabbit (changes requested):
Hey thanks for submitting this!

A major problem here is that you don't actually own this RDNN. `https://github.com/ecommunity` is owned by someone else. The RDNN here should be `io.github.elementary-community.app-generator` since that's where it's hosted. That's why I asked previously if you wanted to make move the org domain to something shorter :)

The screenshot in this release doesn't match the app as released here. I would also recommend versioning this URL so that it pulls the right screenshot, for example, it should be `https://raw.githubusercontent.com/elementary-community/app-generator/1.0.0/data/io.github.ecommunity.app-generator.png`

In metainfo, the `<developer_name>` is deprecated. The `<developer>` tag should be used instead https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer

`<value key="x-appcenter-color-primary-text">` is no longer used. Text color is generated automatically in AppCenter

Your metainfo file is missing a `<content_rating>` tag. This can just be a self-closing tag if there's nothing to disclose https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-content_rating

ryonakano:
> The RDNN here should be `io.github.elementary-community.app-generator` since that's where it's hosted.

Maybe it should be `io.github.elementary_community.app-generator` instead, because Flatpak doesn't allow use of hyphens other than the last element of the RDNN

https://docs.flatpak.org/en/latest/conventions.html#application-ids

danirabbit (approved):
There's still an issue with the screenshot here. It is missing window shadows and appears to be unfocused, but I'll let it through this time since I know you've been waiting on this for a while and I'm sure you'll have more updates soon :)

PR #537, opened 2024-10-31
Update com.github.phase1geo.annotator.json to version 2.0.0.

phase1geo:
See https://github.com/phase1geo/Annotator/releases/tag/2.0.0 for release information.

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime


danirabbit (approved):
Looks great! Congrats on the GTK 4 port!

Just a heads up, LibPortal is included in runtime 8 now so you shouldn't need to also include it yourself :)

PR #540, opened 2024-11-23
Add Sanctify

SubhadeepJasu:
Screenshot from 2024-11-23 21 40 49

## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

### DIFF
--- a/applications/com.github.elliegames.sanctify-game.json
+++ b/applications/com.github.elliegames.sanctify-game.json
@@ -1,5 +1,5 @@
 {
     "source": "https://github.com/elliegames/sanctify-game.git",
-    "commit": "62a8d9886e99847b7b7740ceb4a44a3c2e61d930",
+    "commit": "859ae4bd7e25013ef83ba0a8c2a53b4757f735c5",
     "version": "1.0.0"
   }
--- a/applications/com.github.elliegames.sanctify-game.json
+++ b/applications/com.github.elliegames.sanctify-game.json
@@ -1,5 +1,5 @@
 {
     "source": "https://github.com/elliegames/sanctify-game.git",
-    "commit": "eb2d30d6bb431d4eac69b71a41d2a770da9465d5",
+    "commit": "62a8d9886e99847b7b7740ceb4a44a3c2e61d930",
     "version": "1.0.0"
   }
--- a/applications/com.github.elliegames.sanctify-game.json
+++ b/applications/com.github.elliegames.sanctify-game.json
@@ -1,0 +1,5 @@
+{
+    "source": "https://github.com/elliegames/sanctify-game.git",
+    "commit": "4a79c76a8fadbbe63c815a053e5265321758562b",
+    "version": "1.0.0"
+  }
--- a/applications/com.github.elliegames.sanctify-game.json
+++ b/applications/com.github.elliegames.sanctify-game.json
@@ -1,5 +1,5 @@
 {
     "source": "https://github.com/elliegames/sanctify-game.git",
-    "commit": "4a79c76a8fadbbe63c815a053e5265321758562b",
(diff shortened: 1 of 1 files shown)


zeebok:
Does this work on the final release of OS8 with pipewire as the default? Not sure if that forces any change to the flatpak permissions necessary

zeebok:
In your appdata, in the name tag, you misspelled Sanctify as Santify (though that fits the festive time of year here in the US :P)

Since your oars rating is all none I believe if you'd like you can simply use `<content_rating type="oars-1.1" />` there

zeebok:
You should update the flatpak yml to grab 1.0 instead of 0.9.9!

SubhadeepJasu:
I have been using pipewire for sometime now. I don't have os8 yet but it might work.

Will fix the typo.

I can't set the release to the current one because of how the pipeline works. The flatpak pipeline uses build artifacts from  Godot pipeline uploaded in a prior release. So for flatpak pipeline to work there has to be a release containing the files. Essentially here 0.9.9 is the same as 1.0.0

zeebok:
> I have been using pipewire for sometime now. I don't have os8 yet but it might work.

Okay, mostly just curious if the portals work properly with pipewire. If it works on OS7 it probably works on OS8 as well. And now that I think about it I did swap over to pipewire myself on OS7 and that is what I am testing with lol

> Will fix the typo.

Woo!

> I can't set the release to the current one because of how the pipeline works. The flatpak pipeline uses build artifacts from Godot pipeline uploaded in a prior release. So for flatpak pipeline to work there has to be a release containing the files. Essentially here 0.9.9 is the same as 1.0.0

I see. Okay! I kind of want to poke at things with flatpak to see if there is something a little simpler you can update the manifest to.

SubhadeepJasu:
Fixed the typo.

zeebok (dismissed):

zeebok:
@danirabbit Do we want to worry about passing ARM builds with games, or bypass that requirement?

SubhadeepJasu:
Dani did mention on Discord that arm is still experimental, it's good to have and not an absolute requirement. 

Also currently I dont have the pipeline setup to build arm and I have no way of testing arm builds.

danirabbit (changes requested):
You should not have `  - '--system-talk-name=org.freedesktop.Accounts'` in your flatpak manifest. This is not required for system wide dark mode. Your app should use the portal for this

Is there a reason you have `  - '--filesystem=home'`? That's a pretty broad permission

I'm also really really hesitant to allow a pre-built binary here. We've always required things to be built on our server to make sure the source code you present and we've reviewed is actually what's in the package. This would break our promise of AppCenter apps being auditable

SubhadeepJasu:
At the moment it's not known how to access portals from the Godot Engine, without upstream support for it. So, that system talk is only thing the engine is recognizing.

Building the game in the flatpak pipeline would required either building the engine or having the engine preinstalled in the image. I tried building the Godot engine from source in the flatpak pipeline, but it didn't work. Also, if you want to audit the code, it's all opensource anyway. 

File system home is being used for Godot's save game system. https://docs.godotengine.org/en/stable/tutorials/io/data_paths.html

SubhadeepJasu:
Let me try some ways.

SubhadeepJasu:
I have made some changes to the manifest but first we need to add the Godot base app to the repo for it to work.

SubhadeepJasu:
Updated release signature to one which is compatible with AppCenter flow. Thanks @davidmhewitt .

SubhadeepJasu:
Thanks a lot David! ❤️

PR #429, opened 2022-12-27
com.github.ryonakano.louper 2.0.0

ryonakano:
Commit history: https://github.com/ryonakano/louper/compare/1.0.2...2.0.0

PR #541, opened 2024-11-25
Create com.github.sammarxz.pomerode.json

sammarxz:
pomerode-preview

## Review Checklist

* [ ] App opens
* [ ] Does what it says
* [ ] Categories match

### AppData

* [ ] Name is unique and non-confusing
* [ ] Matches description
* [ ] Matches screenshot
* [ ]  Custom colors meet WCAG A contrast or greater
* [ ] OARS info matches

### Flatpak

* [ ]  Uses elementary runtime
* [ ]  Sandbox permissions are reasonable

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,1 @@
-/.github/ @elementary/core btkostner
 /applications/ @elementary/app-reviewers
--- a/.github/workflows/checklist.yml
+++ b/.github/workflows/checklist.yml
@@ -20,7 +20,7 @@
     
         steps:
           - name: Checklist
-            uses: elementary/action-appcenter-review-checklist@v1.1.0
+            uses: elementary/action-appcenter-review-checklist@v1.2.0
             with:
               token: ${{ secrets.GIT_USER_TOKEN }}
               body: |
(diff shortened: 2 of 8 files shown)


danirabbit (changes requested):
Hey thanks for your submission. It looks like your app was build against the GNOME Runtime. Apps in AppCenter must be build against the elementary runtime. See publishing requirements for this and other requirements:

ryonakano:
@sammarxz The above issue has been resolved by https://github.com/sammarxz/pomerode/pull/1, but you need to create a new release to include these changes and update the `commit` and `version` in `applications/com.github.sammarxz.pomerode.json` in this branch.

Also, your app uses Adwaita stylesheet which is not accepted (see https://github.com/elementary/appcenter-reviews/pull/429#pullrequestreview-1231071137). However, you don't need to subclass Gtk.Application these days and just calling `Granite.init ()` in the `startup ()` method of your application class (like [this](https://github.com/elementary/iconbrowser/blob/e29ba5ea054a48e7d5dd2fba2db4f65fe224e9b8/src/Application.vala#L17)) should work.

PR #544, opened 2024-12-26
Create io.github.ellie_commons.tomato

alainm23:
## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] ~~Custom colors meet WCAG A contrast or greater~~ We no longer check this
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime

image

https://github.com/ellie-commons/tomato

### DIFF
--- a/.github/workflows/checklist.yml
+++ b/.github/workflows/checklist.yml
@@ -36,7 +36,6 @@
                 - [ ] Matches screenshot
                 - [ ] Launchable tag with matching ID
                 - [ ] Release tag with matching version and YYYY-MM-DD date
-                - [ ] Custom colors meet WCAG A contrast or greater
                 - [ ] OARS info matches
                 
                 ### Flatpak
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,7 +45,7 @@ jobs:
        fail-fast: false
 
     container:
-      image: ghcr.io/elementary/flatpak-platform/runtime:7.1-${{ matrix.arch }}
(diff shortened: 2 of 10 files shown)


alainm23:
@elementary/app-reviewers

ryonakano (changes requested):
I'm sorry for late review :bow:

I created the following issues as requested changes:

- https://github.com/ellie-commons/tomato/issues/2
- https://github.com/ellie-commons/tomato/issues/3
- https://github.com/ellie-commons/tomato/issues/4

I'd like you to address them.

ryonakano:
Another thing I'm concerned is the app name: Tomato.

I think your app is fork/continuation of https://github.com/tomatoers/tomato, which was anciently available at AppCenter (maybe in 0.4 Loki). And you inherited the app name, which is potentially a problem if the original author changed their mind, come back to develop the app, and tried publishing the app on AppCenter again. While this wouldn't happen because the original repository has been archived, it's technically possible to unarchive their repository.
So I'm not sure if this is fine or not good.

@danirabbit Should we choose a different name for a forked app from the original app which is previously available on AppCenter?

danirabbit:
I think since it's archived and the original author hasn't tried to publish in all these years that's probably okay

ryonakano (approved):
LGTM.

@elementary/app-reviewers I'm new to review app submissions. Feel free to review if I'm missing something.

PR #557, opened 2025-02-24
Create io.github.ellie-commons.teamcons.jorts.json

teamcons:
## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime


zeebok (dismissed):

zeebok:
@danirabbit I haven't seen this failure to find a manifest failure. Is this not finding something in the app's repo? The log doesn't really give any info.

teamcons:
> @danirabbit I haven't seen this failure to find a manifest failure. Is this not finding something in the app's repo? The log doesn't really give any info. Does the flatpak file _need_ to be a JSON file?

could be it faceplants because the extension is .yaml and not .yml

danirabbit:
@teamcons probably haha

danirabbit (changes requested):
While we're making another release, the "developer_name" tag in metainfo is deprecated and should be replaced: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer

Also your meta info should not reference elementary or elementary OS. Every app on AppCenter is made for elementary OS and your app info page will already have a "made for elementary OS" badge

 The "x-appcenter-color-primary-text" tag is unused and "x-appcenter-color-primary" is deprecated. You should instead use the Branding tag: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-branding

danirabbit (changes requested):
Your usage of the `developer` tag is incorrect. Please check the docs :) It should look something like this instead:

```
  <developer id="io.github.teamcons">
    <name>Stella - Teamcons</name>
  </developer>
```

I just noticed your screenshot as well. It should contain only your application's window. It shouldn't include your wallpaper. See: https://docs.elementary.io/develop/appcenter/publishing-requirements#metainfo-and-oars

teamcons:
Okay. Ive taken into account your feedback and made some more improvements.
This is looking super super serious and clean now >:3
I think end of the week or next week i will do a translation update and maybe a screenshot or two, but for now this needs to be out.

teamcons:
The desktop comment didnt match the description. Shoddy work.
Now this should be enough.

danirabbit (approved):
Looks awesome, great work!

PR #564, opened 2025-04-05
Update Jorts to 2.4.2

teamcons:
## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/io.github.ellie_commons.jorts.json
+++ b/applications/io.github.ellie_commons.jorts.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ellie-commons/Jorts.git",
-  "commit": "649e0a12dadecd136aaa9f9d7d0c29472cdb8538",
-  "version": "2.4.1"
+  "commit": "922730da20db631e8541976a094280967507b344",
+  "version": "2.4.2"
 }
--- a/applications/io.github.ellie_commons.jorts.json
+++ b/applications/io.github.ellie_commons.jorts.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ellie-commons/Jorts.git",
-  "commit": "f7dbb1c56b4a9ec8874e91cf3cc180920ac8e77e",
-  "version": "2.4.0"
+  "commit": "649e0a12dadecd136aaa9f9d7d0c29472cdb8538",
+  "version": "2.4.1"
 }
--- a/applications/io.github.ellie_commons.jorts.json
+++ b/applications/io.github.ellie_commons.jorts.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ellie-commons/Jorts.git",
-  "commit": "f1612e38b139cd3760d4383bb940f9097c09b9a4",
-  "version": "2.3.0"
+  "commit": "f7dbb1c56b4a9ec8874e91cf3cc180920ac8e77e",
+  "version": "2.4.0"
 }


ryonakano:
You set May 4 not April 4 as the release date. Is this expected?

https://github.com/ellie-commons/Jorts/blob/f7dbb1c56b4a9ec8874e91cf3cc180920ac8e77e/data/io.github.ellie_commons.jorts.metainfo.xml.in#L90

teamcons:
No. Oversight.

ryonakano:
Just out of curiosity (I don't mean to block your release), but is there any reason why you removed the close button in the headerbar? I personally thought now it looks as if there is no visual widgets to save notes and close the app, while it's still possible with Ctrl+Q, Alt+F4, "Close" menuitem in the context menu of the headerbar, etc. though.

image

Apparently you looks like changed this intentionally in 2.4.0:

```diff
[ryo@b760m ~/work/tmp/Jorts (main =)]$ git diff 2.3.0..2.4.0 -- src
:

diff --git a/src/MainWindow.vala b/src/MainWindow.vala
index 7360de7..80d5b1e 100644
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
:

             /**********************************************/
             /*              USER INTERFACE                */
             /**********************************************/

-            var headerbar = new Gtk.HeaderBar();
+            this.headerbar = new Gtk.HeaderBar();
             headerbar.add_css_class ("flat");
             headerbar.add_css_class("headertitle");
             //header.has_subtitle = false;
-            headerbar.set_show_title_buttons (true);
-            headerbar.decoration_layout = "close:";
+
+            //headerbar.decoration_layout = "close:";
+            headerbar.set_show_title_buttons(false);
```

teamcons:
> Just out of curiosity (I don't mean to block your release), but is there any reason why you removed the close button in the headerbar? I personally thought now it looks as if there is no visual widgets to save notes and close the app, while it's still possible with Ctrl+Q, Alt+F4, "Close" menuitem in the context menu of the headerbar, etc. though.

There was a lot of back and forth on my side about this, with two competing approaches:
-provide a way to reopen all notes, including closed ones, with a desktop action
-follow through on the minimalist approach: the idea was that if you do not need a note anymore, you simply delete it.

i went with the latter - it does not really make sense to manage notes between closed but still needed and visible ones.

Should the user really need to note down something, but also keep it out of sight like until next login (when all of the storage is loaded again), there is still the right-click -> Close or the overview with all notes when you click on the icon.
The notes are saved all the same.

I didnt provide a toggle to let user choose. I would eventually be open to if people have a workflow depending on opening/closing notes, so they can have a close and desktop action for reopen all,
But id like, like eOS, to provide a one flow approach, and avoid featurecreep - not try to do too much, not overwhelm too much

ryonakano:
Thanks for clarification. Users might want to close a private note without deleting it when they want to hide it temporary, but your app already covers that case with another approach. So maybe removing the close button wouldn't be a big problem.

Anyway, I don't mean to block your release anymore as I said :) Approving!

ryonakano (approved):
Works confirmed with 2.4.2 and no concerned changes are found between 2.3.0 and 2.4.2

PR #574, opened 2025-05-05
Create com.github.eppixx.buoy

eppixx:


applications/com.github.eppixx.buoy
```diff
@@ -0,0 +1,5 @@
+{
+  "source": "https://github.com/eppixx/buoy.git",
+  "commit": "1aa1ab3449e6c7e86e93dc127a96b8ce786f67e0",
+  "version": "v1.0.0"
```
danirabbit, line 4: This needs to be semantic version without any letters:
```suggestion
  "version": "1.0.0"
```

danirabbit (changes requested):
Thanks for your app submission! It's not currently building because of the version tag I think, but I also have some other comments:

Your appdata seems to be missing a [developer](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer) tag

I'd also recommend adding more [url](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url) tags such as a link to your issue tracker, etc

In your screenshots, instead of [doing this](https://raw.githubusercontent.com/eppixx/buoy/develop/data/screenshots/dashboard_dark_light.png) you can specify if your screenshot is light mode or dark mode. I would highly recommend doing that instead. Check out the "environment" property specifically: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

You can also only have one `default` screenshot. It's not valid to have all of your screenshots marked as `default` :)

Your screenshot should also not be maximized. It should have a shadow and a transparent background. I recommend making your app window as small as possible while still showing its contents in your screenshot. Check out some of the other apps in AppCenter as an example of how your screenshots will appear. If your screenshot is too large it will be hard for people to see what's in your app when it's shrunk down to thumbnail sizes

I'd highly recommend adding a [branding](https://docs.elementary.io/develop/writing-apps/our-first-app/metadata#branding) tag with your app's brand colors. This really helps your app to stand out and look nice in banners and makes your app's info page feel more personal

Granite has automatic stylesheet loading built-in! So if you rename your css file to `Application.css` or alias it [here](https://github.com/eppixx/buoy/blob/1aa1ab3449e6c7e86e93dc127a96b8ce786f67e0/src/resources.gresource.xml.in#L5) it will be loaded automatically and you don't have to add it yourself [here](https://github.com/eppixx/buoy/blob/1aa1ab3449e6c7e86e93dc127a96b8ce786f67e0/src/main.rs#L112)

--

I tried to build your app locally, but it error'd out. I'm not personally familiar with writing apps in Rust so I can't personally offer any good advice, but if the build fails in CI after the versioning is fixed, maybe someone in discord can help out: https://discord.gg/pCenUBu

eppixx:
Thanks for taking your time with the submission.

> Your appdata seems to be missing a [developer](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer) tag

done

> I'd also recommend adding more [url](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url) tags such as a link to your issue tracker, etc

added more url tags

> In your screenshots, instead of [doing this](https://raw.githubusercontent.com/eppixx/buoy/develop/data/screenshots/dashboard_dark_light.png) you can specify if your screenshot is light mode or dark mode. I would highly recommend doing that instead. Check out the "environment" property specifically: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

> You can also only have one default screenshot. It's not valid to have all of your screenshots marked as default :)

> Your screenshot should also not be maximized. It should have a shadow and a transparent background. I recommend making your app window as small as possible while still showing its contents in your screenshot. Check out some of the other apps in AppCenter as an example of how your screenshots will appear. If your screenshot is too large it will be hard for people to see what's in your app when it's shrunk down to thumbnail sizes

added new screenshots. By using the environment tag my ci fails at the metadata validation step. I don't know if the version of the validation binary is not recent enough or if it is something else, so I removed them for now. Also removed the duplicated default tag.

> I'd highly recommend adding a [branding](https://docs.elementary.io/develop/writing-apps/our-first-app/metadata#branding) tag with your app's brand colors. This really helps your app to stand out and look nice in banners and makes your app's info page feel more personal

added a branding color

> I'd highly recommend adding a [branding](https://docs.elementary.io/develop/writing-apps/our-first-app/metadata#branding) tag with your app's brand colors. This really helps your app to stand out and look nice in banners and makes your app's info page feel more personal

I'm not loading the css from the environment. I'm including it in the binary when compiling. If this is a dealbreaker I can change that.

> I tried to build your app locally, but it error'd out. I'm not personally familiar with writing apps in Rust so I can't personally offer any good advice, but if the build fails in CI after the versioning is fixed, maybe someone in discord can help out: https://discord.gg/pCenUBu

If the error persists, please include the error in your reply.

danirabbit:
@eppixx I think maybe this isn't building in CI because the file is named with `.yaml` and CI is expecting either `.yml` or `.json`

davidmhewitt:
@eppixx You will need to generate your `cargo-sources.json` file and check it into the repository. It should be treated like your Cargo lockfile as it locks the flatpak dependency versions to each commit in your repo.

You could even add a GitHub action to update it every time your Cargo lockfile updates:
https://github.com/davidmhewitt/elementary-rust-example/blob/main/.github/workflows/flatpak-modules.yml

eppixx:
"rustc" not found. seems like the sdk-extension `org.freedesktop.Sdk.Extension.rust-stable` is not used

PR #575, opened 2025-05-07
Create io.github.wpkelso.slate.json

wpkelso:
## Review Checklist

- [x] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [x] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/io.github.wpkelso.slate.json
+++ b/applications/io.github.wpkelso.slate.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/wpkelso/slate.git",
+  "commit": "92f3db475907ada3abdc3e2005f27aa99cc691b9",
+  "version": "1.0.0"
+}
--- a/applications/io.github.wpkelso.slate.json
+++ b/applications/io.github.wpkelso.slate.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/wpkelso/slate.git",
-  "commit": "92f3db475907ada3abdc3e2005f27aa99cc691b9",
-  "version": "1.0.0"
+  "commit": "e9c807a8c8deba0acc6e27e202d565943dd5153a",
+  "version": "1.0.1"
 }


danirabbit (changes requested):
Congrats on your first app! It's so good it makes me want to cry. You did everything like almost perfectly, best practice, so good. Honestly a joy to see your submission. I'm so proud 🩷

**There's just a couple of blockers:**

Looks like a copy paste error, you appdata summary says "Mail" :) https://github.com/wpkelso/slate/blob/92f3db475907ada3abdc3e2005f27aa99cc691b9/data/slate.metainfo.xml.in#L9

Your screenshot URL is returning a 404 currently

**This stuff is optionally advice:**

AppCenter can show the correct light or dark mode screenshot automatically instead of doing [this slashy thing](https://raw.githubusercontent.com/wpkelso/slate/1.0.0/data/screenshot.png) you should supply a separate dark and light mode screenshot and mark them with the environment tag: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

I don't think your app actually needs `--filesystem=xdg-documents` since it uses the file portal to open and save files. You might want to double check if you can remove this permission

In your headerbar, there's actually a property to show window controls so you don't have to add them manually: https://valadoc.org/gtk4/Gtk.HeaderBar.show_title_buttons.html

In other apps with a split view we pack the headebar into a box manually because by default it goes across the whole top of the window. In your app, since you only have one view, you can just set your custom header as the titlebar instead of doing the blank titlebar trick. Unless you're intending to add a sidebar or something, then ignore me :)

[This](https://github.com/wpkelso/slate/blob/92f3db475907ada3abdc3e2005f27aa99cc691b9/src/Window.vala#L98) seems to be just copy/paste from Code? I don't think your app will actually have permission to do this

teamcons:
I will definitively be using this, mister otter :)

danirabbit (changes requested):
One more small thing! It looks like your screenshot URLs are 404ing because you tagged your release "v1.0.1" instead of "1.0.1". So if you update the tag we should be good to go :)

wpkelso:
@danirabbit Tags should be good to go now

danirabbit (approved):
Fantastic work! Congrats :tada:

PR #580, opened 2025-06-14
Create com.github.stsdc.gadget_clock.json

stsdc:
## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/com.github.stsdc.gadget_clock.json
+++ b/applications/com.github.stsdc.gadget_clock.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/stsdc/gadget_clock.git",
-  "commit": "1a88285a18fa6bc74cb5cfb621ba98e4cb9c48f3",
-  "version": "0.0.3"
+  "commit": "2eb072bd0685319a82d90b9a293399810548107b",
+  "version": "0.4.0"
 }
--- a/applications/com.github.stsdc.gadget_clock.json
+++ b/applications/com.github.stsdc.gadget_clock.json
@@ -1,0 +1,5 @@
+{
+  "source": "https://github.com/stsdc/gadget_clock.git",
+  "commit": "1a88285a18fa6bc74cb5cfb621ba98e4cb9c48f3",
+  "version": "0.0.3"
+}


ryonakano (changes requested):
## Requirements
- [ ] Your manifest [must use](https://docs.elementary.io/develop/appcenter/publishing-requirements#packaging) the elementary runtime.
- [ ] Make sure your app can build without internet connection. I couldn't build your app even if I install deps from Flathub.
- [ ] Your metainfo file fails to validate.
```
user@elementary-8-daily:~/work/gadget_clock$ appstreamcli validate /usr/share/metainfo/com.github.stsdc.gadget_clock.metainfo.xml 
E: com.github.stsdc.gadget_clock:33: web-url-expected 
W: com.github.stsdc.gadget_clock:34: url-not-reachable
     https://github.com/stsdc/gadget_clock/discussions - URL was not found on the server.
E: com.github.stsdc.gadget_clock:~: app-description-required

\u2718 Validation failed: errors: 2, warnings: 1, pedantic: 1
user@elementary-8-daily:~/work/gadget_clock$
```
- [ ] Out of curiosity, but why you set `social-info` OARS data to mild in [your metainfo file](https://github.com/stsdc/gadget_clock/blob/1a88285a18fa6bc74cb5cfb621ba98e4cb9c48f3/data/gadget_clock.metainfo.xml.in#L30)?

## Suggestions
- I personally recommend you to use the `io.github.*` naming notation instead of `com.github.*`. The latter notation is no longer accepted on Flathub, while we do, so you would get bothered if you try to submit your app to Flathub too.
- Inconsisten version. You set version to 0.3.0 [in your meson.build](https://github.com/stsdc/gadget_clock/blob/1a88285a18fa6bc74cb5cfb621ba98e4cb9c48f3/meson.build#L2) and [in your metainfo file](https://github.com/stsdc/gadget_clock/blob/1a88285a18fa6bc74cb5cfb621ba98e4cb9c48f3/data/gadget_clock.metainfo.xml.in#L48-L55) but the git tag says 0.0.3.
- The `Comment` in [your desktop file](https://github.com/stsdc/gadget_clock/blob/1a88285a18fa6bc74cb5cfb621ba98e4cb9c48f3/data/gadget_clock.desktop.in#L3) is noun, but we recommend to follow [this formula](https://docs.elementary.io/hig/app-launcher#:~:text=Name%20is%20a(n)%20GenericName%20that%20helps%20you%20Comment.)
- I suppose your app has leftovers from some examples. The macro names you used in your include guards say `GTKMM_EXAMPLE` or `MYAREA` which sounds just copied & pasted from examples in some docs. Also, there is [a signal handler](https://github.com/stsdc/gadget_clock/blob/1a88285a18fa6bc74cb5cfb621ba98e4cb9c48f3/src/clock.cpp#L36) that just prints Hello World.
- It is not recommended to use Gdk::Pixbuf to think about porting to GTK5 in the future. See: https://docs.gtk.org/gtk4/migrating-4to5.html#stop-using-gdkpixbuf
- You should describe your app in more detail in the README, e.g. how to build your app, screenshot(s), etc.

PR #583, opened 2025-06-28
Update Reminduck (it has been so long!)

teamcons:
A new version, with modern port, of Reminduck

This is not just a fork of [the old unmaintained one](https://github.com/elementary/appcenter-reviews/blob/main/applications/com.github.matfantinel.reminduck.json), but its continuation transferred from Matt (i can share the email exchange)

would the old one need to be deleted? there risks to be confusion.
or would it work to softlink the old one to this one?

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/com.github.matfantinel.reminduck.json
+++ b/applications/com.github.matfantinel.reminduck.json
@@ -1,5 +1,7 @@
 {
   "source": "https://github.com/matfantinel/reminduck.git",
   "commit": "b66efd6657ed74da38be5ba94a8e3ef51f651989",
-  "version": "1.6.2"
+  "version": "1.6.2",
+  "end-of-life": "Updated by the ellie-commons initiative",
+  "end-of-life-rebase": "io.github.ellie_commons.reminduck"
 }
--- a/applications/com.github.matfantinel.reminduck.json
+++ b/applications/com.github.matfantinel.reminduck.json
@@ -1,7 +1,5 @@
 {
   "source": "https://github.com/matfantinel/reminduck.git",
   "commit": "b66efd6657ed74da38be5ba94a8e3ef51f651989",
-  "version": "1.6.2",
-  "end-of-life": "Updated by the ellie-commons initiative",
-  "end-of-life-rebase": "io.github.ellie_commons.reminduck"
+  "version": "1.6.2"
 }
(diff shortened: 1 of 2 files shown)


danirabbit (changes requested):
Hey, so yes you need to add an `eol` line to [the old Reminduck .json](https://github.com/elementary/appcenter-reviews/blob/main/applications/com.github.matfantinel.reminduck.json) like [this one](https://github.com/elementary/appcenter-reviews/pull/406/files).

I can see that trying to navigate to the URL for reminduck redirects to ellie-commons so that's good enough evidence for me! Thanks for maintaining this app :)

---

You should definitely not add the session bus to your sandbox holes. You've done something wrong here if you think you need `--socket=session-bus`. The most common issue folks have here is they've got some app ID mismatch

Likewise, portals are already automatically part of the sandbox exceptions. You never need to manually add a sandbox hole to use portals so you should not add `talk-name=org.freedesktop.portal.Background`

---

In metainfo, looks like a copy/paste error:
`<translation type="gettext" source_locale="en_GB">io.github.ellie_commons.jorts</translation>`

Should this not be `io.github.ellie-commons`?
`developer id="io.github.teamcons`

Not a blocker but for a future version, you should mark your dark screenshots properly with an environment tag: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

teamcons:
yes im on it ma'am

teamcons:
> Should this not be `io.github.ellie-commons`? `developer id="io.github.teamcons`

Im not sure. There's a project group tag, so i assumed this specific one honours the individual behind.
Appstream isnt super clear for both

teamcons:
Appcenter complains because of the names?...
This one should take your feedback into account - and i got two bugs out in the process :)
there is still a lot of work under the hood...

danirabbit:
Just looking at your changes here and `cmd.add ("com.github.elfenware.reminduck");` in Application.vala is definitely not right :sweat_smile: To avoid issues like this you could use the [`application_id`](https://valadoc.org/gio-2.0/GLib.Application.application_id.html) variable that you set in your Application class

I would also recommend not making the screenshots larger. It's much better to have screenshots be with a window that is as small as is reasonable so that they aren't super shrunk down when viewed in AppCenter

But neither are publishing blockers for me. Happy to get this merged and looking forward to future releases :)

danirabbit (approved):

PR #588, opened 2025-07-01
Update io.github.libredeb.hashit.json to 0.1.4

libredeb:
Bump to version 0.1.4

### DIFF
--- a/applications/io.github.libredeb.hashit.json
+++ b/applications/io.github.libredeb.hashit.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/libredeb/hash-it.git",
-  "commit": "d17437b00b8949dea73420aceaa842bb895222da",
-  "version": "0.1.3"
+  "commit": "bad4ac006a79208c54a0c9575c90ba7acaa58a2b",
+  "version": "0.1.4"
 }


libredeb:
@ryonakano wait a moment, maybe I should replace the parameter by this other one that is more “restrictive” and logical to use in my application:
```yaml
finish-args:
  - '--filesystem=home'
```

If this fix is correct I can work on “cleaning up” the released version 0.1.4 so that it is in good condition before proceeding. How do you see it?

ryonakano:
@libredeb Yeah, that's no problem as a temporary workaround (maybe dragging from an external storage would cause the issue, but most users wouldn't do that)

ryonakano:
@libredeb Also don't forget to add release notes and bump your project versions.

libredeb:
@ryonakano thanks, regarding to version number increment, I think it's best to do them when I'm including new functionality, fixes, or something else worthwhile, right? For now, to make sense of everything, I'm "remediating" a specific release (0.1.4 in this case).

ryonakano:
> regarding to version number increment, I think it's best to do them when I'm including new functionality, fixes, or something else worthwhile, right?

You modified the Flatpak manifest in the new `commit` revision of https://github.com/elementary/appcenter-reviews/pull/588/commits/07760ffe6b1de235889b9a0c569729137b43e13e as I requested you to do so, which is a *fix*, so I expected you to create a new release on your repository (like 0.1.5) instead of deleting and recreating the 0.1.4 release/tag that you'd already published.

We expect developers to increment version number each time when you modify the source code even if during the review process. IIRC we don't force this but it's a best practice to guarantee reproducible builds for communities. FYI [our old blog post](https://blog.elementary.io/developer-tips-github-workflow/) says:

> While GitHub doesn’t treat releases as immutable, it’s best practice for you to. Once you make a release, other contributors can start downloading tarballs and packaging them elsewhere…… If you modify a release, you’re bound to cause issues. It’s best to make a small x.y.1 patch release if you need to modify something, and be clear in the release description on GitHub what was modified or fixed.

However, the old 0.1.4 release/tag has gone already so I'm fine to release this as 0.1.4 for this time.

libredeb:
@ryonakano I understand you well, thanks for the patience and the corresponding explanations. This is a learning journey for me about application development and I appreciate all your help.

However, I am uncomfortable with how the `appdata.xml` file is treated as the version increases, it is something that I should also update/translate at each iteration. It confuses me there what is the workflow for polishing a release before it is approved. What would be the right way?

For now let's release this version as `0.1.4` and I will take into consideration these best practices for the next release. Thank you and big hug!

ryonakano:
> However, I am uncomfortable with how the `appdata.xml` file is treated as the version increases, it is something that I should also update/translate at each iteration. It confuses me there what is the workflow for polishing a release before it is approved. What would be the right way?

We're sorry for the inconvenience. For example, inviting someone who is familiar with app development on AppCenter to your repository and asking them for review before releasing or in each change would increase possibility to publish your app without rejection.

PR #589, opened 2025-07-03
Update com.github.elfenware.badger to v5.0.0

dar5hak:
Badger is now v5! :tada:

The biggest changes are:
- It's now GTK4
- It uses the background portal to autostart instead of manually installing a .desktop file

And there are some nice UI updates as well.

Most of this is thanks to @teamcons. They're awesome.

### DIFF
--- a/applications/com.github.elfenware.badger.json
+++ b/applications/com.github.elfenware.badger.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/elfenware/badger.git",
-  "commit": "5ff6559e85943d8dd6a368df7c3436aafe394bf2",
-  "version": "4.4.0"
+  "commit": "6d5eaac8e51f4b67b2ce4fff3353534cb4747e80",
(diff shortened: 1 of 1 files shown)


teamcons:
Hi Ryo!
The flathub manifest is a draft i added at some point. It is half baked - and would either be deleted or worked upon depending if theres interest to have badger there as well.

ryonakano:
> The flathub manifest is a draft i added at some point. It is half baked - and would either be deleted or worked upon depending if theres interest to have badger there as well.

That's fine, I just wrote the suggestion as FYI in case you two might submit this manifest as it is for review on Flathub.

But it looks like previously Badger published on Flathub but archived soon? https://github.com/flathub/flathub/pull/2454#issuecomment-902717344

Anyway, whether to publish an app on Flathub too or not is up to its maintainers and you should discuss with @dar5hak in the repository of Badger.

PR #590, opened 2025-07-06
Update jorts to 3.3.0

teamcons:
Reminduck will follow very soon
We've revived Picker btw: https://github.com/ellie-commons/cherrypick
But we cant contact the dev, so this is waiting for a new icon (we have someone new who wanna contribute :) )


ryonakano (approved):
This would be pedantic though, I suppose you shouldn't remove the name of the original author, Lains, from the copyright holder text if the source code is originally written by them and not written by you from scratch (I think the copyright header is not a thing that can easily "unify").

But this is out of scope of the review process so approving.

danirabbit:
@teamcons I highly recommend this free course from Linux Foundation about software licensing basics. It will explain how to use copyright and license headers etc https://training.linuxfoundation.org/training/open-source-licensing-basics-for-software-developers/

PR #592, opened 2025-07-06
Update Reminduck to 2.1.0

teamcons:
## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/io.github.ellie_commons.reminduck.json
+++ b/applications/io.github.ellie_commons.reminduck.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ellie-commons/reminduck.git",
-  "commit": "c7f983f34cc47d19049565df7d7adfdaf50f5529",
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Looks good! Nice update

Couple things for a future update:
* Your menu button doesn't have a tooltip. Image buttons should always have a tooltip or they will be unlabeled for the screen reader
* You can/should set [`primary`](https://valadoc.org/gtk4/Gtk.MenuButton.primary.html) on your menubutton so that it gets the standard main menu shortcut

teamcons:
Noted. Stella intended to move the popover into a stack page. But knowing how to accessibility is very welcome.
We will however leave programmation aside for a bit.

PR #603, opened 2025-07-23
Update Cherrypick to 2.1.0

teamcons:
Ryo's nitpick on a missing space made me look for an opportunity for a new release, because it is the kind of detail you cant unsee

Their PR kicked me into taking care of a deprecated widget and doing finally a release with all the polish. If Cherrypick ends up unmaintained at least it can hold a solid while now

(and now i realize i left it at 8 instead of 8.2... Hope theres not too much missing out)

## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime

### DIFF
--- a/applications/io.github.ellie_commons.cherrypick.json
+++ b/applications/io.github.ellie_commons.cherrypick.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ellie-commons/cherrypick.git",
-  "commit": "f90ac45ca2ed07eb48939269da5332308635e94c",
-  "version": "2.0.1"
+  "commit": "33cb31facb513fd0ab28786d0e1da9d0724287d9",
+  "version": "2.1.0"
 }


ryonakano (approved):
The release date of 2.1.0 seems to be just copied and pasted from 2.0.1. I won't block publishing your release this time though, because there are no other problems in 2.1.0 and it's foolish to let you create another release to address this.

```diff
+    <release date="2025-07-14" version="2.1.0">
```

ryonakano:
> (and now i realize i left it at 8 instead of 8.2... Hope theres not too much missing out)

It's fine if you're using 8.x version of the platform at the moment. Note that using 8.2 at the moment would encounter the crash I fixed in https://github.com/elementary/granite/pull/893

PR #609, opened 2025-09-13
Update cherrypick to 2.1.2

teamcons:
One simple bugfix where the desktop action did not work if the app was already open

### DIFF
--- a/applications/io.github.ellie_commons.cherrypick.json
+++ b/applications/io.github.ellie_commons.cherrypick.json
@@ -1,5 +1,5 @@
 {
   "source": "https://github.com/ellie-commons/cherrypick.git",
-  "commit": "33cb31facb513fd0ab28786d0e1da9d0724287d9",
-  "version": "2.1.0"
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Looks good 🚀 

---

I noticed you removed the `Version` key from the .desktop file. I would recommend putting that back. If anything it would be good to update it to the latest version, but that lets the system know which version of the spec your launcher is targeting: https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html

teamcons:
> Looks good 🚀
> 
> I noticed you removed the `Version` key from the .desktop file. I would recommend putting that back. If anything it would be good to update it to the latest version, but that lets the system know which version of the spec your launcher is targeting: https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html

Ah noted

Speaking of . Desktop files i wanted to ask if eOS supports the Dynamic Launcher portal?

danirabbit:
I'm fairly certain we do. I think web depends on it for web apps. If not please open an issue report :)

PR #664, opened 2026-04-24
Update Jorts to 4.1.1

teamcons:
## Review Checklist

- [ ] App opens
- [ ] Does what it says
- [ ] Categories match

### AppData
- [ ] Matches description
- [ ] Matches screenshot
- [ ] OARS info matches

### Flatpak
- [ ] Uses elementary runtime


teamcons:
bumped Granite-7 to 7.7.0 to take care of the deprecations, but it seems too early for that for whats available in appcenter (7.6.0)

teamcons:
there wasnt much different so we undid the 7.7.0 stuff

EDIT: Latte theme does not build. Reverting, and will use a flag.
So people will be able to get the pride variant in June.

ryonakano (approved):

PR #673, opened 2026-05-05
Update Flatpak repository URLs in publish.yml

davidmhewitt:
New `flat-manager` server release saves screenshots at https://flatpak.elementaryos.org/repo/media/ now instead of https://flatpak.elementaryos.org/repo/screenshots/

Update the base URL so the published metadata points to the right path.

Update to the .org domain while we're here

@ryonakano This is why the screenshot doesn't load in the latest release of Reco (https://appcenter.elementary.io/com.github.ryonakano.reco/), but we can republish after merging this and it should fix it.


ryonakano (approved):
Thanks!

PR #687, opened 2026-07-14
Submit com.github.gylangsatria.clipstory v1.5.1

gylangsatria:
### Description
A lightweight clipboard history manager built for elementary OS using GTK and Granite. Automatically stores copied text and allows quick search, reuse, and management of clipboard items.

### Features
- Automatic clipboard monitoring (X11 & Wayland)
- Searchable clipboard history with pagination
- Pin/unpin important items
- Dark mode support
- Configurable history limit
- Double-click to copy
- Auto-start at login option

### Screenshots
Screenshot Dark
Screenshot Light

## Review Checklist

- [x] App opens
- [x] Does what it says
- [x] Categories match

### AppData
- [x] Matches description
- [x] Matches screenshot
- [x] OARS info matches

### Flatpak
- [x] Uses elementary runtime


applications/com.github.gylangsatria.clipboard-history.json
```diff
@@ -0,0 +1,5 @@
+{
+  "source": "https://github.com/gylangsatria/clipboard-history-elementaryos",
+  "commit": "16412a95622128bdf57389470a4d7f7d684861c3"
```
ryonakano:
```suggestion
  "commit": "16412a95622128bdf57389470a4d7f7d684861c3",
```

ryonakano:
The JSON file in your source code is not as the same as the one you submit into this repository.

https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.json

It must be the Flatpak manifest but I don't think you need it anyways because you seem to already have the Flatpak manifest as the YAML format:

https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml

ryonakano (changes requested):
I've just take a look at the Flatpak manifest, but leaving some changes requests and questions:

Changes requests:

- My previous comment at https://github.com/elementary/appcenter-reviews/pull/687#issuecomment-4997318202
- Please use `--socket=x11-fallback` instead of `--socket=x11` when you use `--socket=wayland`
  - https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L9
- I think you can send notifications without `--talk-name=org.freedesktop.Notifications`
  - https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L14

Questions:

- Is there any reason you didn't use the latest `8.x` version of the runtime?
  - https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L3
- Is `--socket=session-bus` really needed? This means to request arbitrary access to the session bus, which is not acceptable
  - https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L12
- What is `--talk-name=org.gtk.Settings` for?
  - https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L16
- Is there any reason you use the external xdg data directory instead of under `~/.var/app/<app_id>` which can be accessible without any additional permission?
  - https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L18

gylangsatria:
> The JSON file in your source code is not as the same as the one you submit into this repository.
> 
> https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.json
> 
> It must be the Flatpak manifest but I don't think you need it anyways because you seem to already have the Flatpak manifest as the YAML format:
> 
> https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml

The JSON file in my source code was initially created as a reference for the submission to this repository. I've already removed it from the source repo since it's not part of the application itself. The Flatpak manifest (.yml) is the correct one.

gylangsatria:
> I've just take a look at the Flatpak manifest, but leaving some changes requests and questions:
> 
> Changes requests:
> 
> * My previous comment at [#687 (comment)](https://github.com/elementary/appcenter-reviews/pull/687#issuecomment-4997318202)
> * Please use `--socket=x11-fallback` instead of `--socket=x11` when you use `--socket=wayland`
>   
>   * https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L9
> * I think you can send notifications without `--talk-name=org.freedesktop.Notifications`
>   
>   * https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L14
> 
> Questions:
> 
> * Is there any reason you didn't use the latest `8.x` version of the runtime?
>   
>   * https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L3
> * Is `--socket=session-bus` really needed? This means to request arbitrary access to the session bus, which is not acceptable
>   
>   * https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L12
> * What is `--talk-name=org.gtk.Settings` for?
>   
>   * https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L16
> * Is there any reason you use the external xdg data directory instead of under `~/.var/app/<app_id>` which can be accessible without any additional permission?
>   
>   * https://github.com/gylangsatria/clipboard-history-elementaryos/blob/v1.4.5/com.github.gylangsatria.clipboard-history.yml#L18

## Review Responses

### 1. Why aren't you using runtime 8.x?

**Answer:**  
I initially used runtime 7 because it was the standard choice when I first set up the project. However, the application has now been updated to runtime 8. 

### 2. Is `--socket=session-bus` really necessary?

**Answer:**  
No. This permission grants access to the entire D-Bus session bus, which is much broader than what the application actually requires. The application only accesses the clipboard through `Gtk.Clipboard`, which relies on the windowing system (X11/Wayland), not the D-Bus session bus. Therefore, this permission was unnecessary and has been removed. 

### 3. What is `--talk-name=org.gtk.Settings` used for?

**Answer:**  
I originally believed this permission was required for dark mode support (`gtk_application_prefer_dark_theme`). However, GTK settings are handled internally by the application and do not require external D-Bus access. As a result, this permission was unnecessary and has been removed. 

### 4. Why use `--filesystem=xdg-data/clipboard-history` instead of the default Flatpak sandbox?

**Answer:**  
The application uses `Environment.get_user_data_dir()` in `clipboard-history.vala`. Inside a Flatpak sandbox, this function automatically resolves to `~/.var/app/com.github.gylangsatria.clipboard-history/data/`, meaning the application's data is already stored in its private Flatpak data directory without requiring any additional filesystem permissions. Therefore, the `--filesystem=xdg-data/clipboard-history` permission was unnecessary and has been removed. 

### 5. Why use `--socket=x11` instead of `x11-fallback`?

**Answer:**  
Since the application also requests `--socket=wayland`, the recommended configuration is to use `--socket=x11-fallback`. This ensures Wayland is used whenever available, while X11 is only enabled as a fallback for systems that do not support Wayland. The manifest has been updated accordingly. 

### 6. Is `--talk-name=org.freedesktop.Notifications` really necessary?

**Answer:**  
No. The application does not currently send desktop notifications, so this permission is not required. It has been removed from the manifest. 


Thank you for taking the time to thoroughly review my submission and provide detailed feedback. I genuinely appreciate your suggestions, as they helped identify unnecessary permissions and improve the application's adherence to Flatpak best practices.

ryonakano (changes requested):
I'm sorry for my late review. Leaving additional five changes requests, a recommendation, and a question.

Changes requests:

- Your application name `Clipboard History` is too generic. Consider renaming to a unique name, e.g. `ClipStory` (a simple coined word of clipboard and history).
  - https://docs.elementary.io/develop/appcenter/publishing-requirements#naming
- You shouldn't refer your application is for elementary OS because it is assumed that all apps submitted to AppCenter are designed for elementary OS
  - https://docs.elementary.io/develop/appcenter/publishing-requirements#metadata-requirements
- You should not use the `-elementaryos` suffix in your repository name `clipboard-history-elementaryos`
  - https://docs.elementary.io/develop/appcenter/publishing-requirements#naming
- I think `com.github.gylangsatria.clipboard-history-autostart.desktop` wouldn't work as expected
  - The correct way that supports Flatpak is to use Portals
  - See https://github.com/elementary/mail/pull/882/changes for example of how to do it (but note that `libportal` is already included in the elementary Flatpak runtime and you don't need to add it in your Flatpak manmifest
- Please make sure to create a new tag and update not only the `commit` tag but also the `version` tag in the JSON file of this branch when there are changes in your source code

Recommendations:

- Now that it is recommended to use `io.github` prefix instead of `com.github` because the latter is reserved for GitHub itself in Flathub and thus your app may not be accepted if you want to submit it there
  - https://docs.flathub.org/docs/for-app-authors/requirements#application-id
- The application icon in the about dialog is missing in my installation, even if I restart my system. I would recommend you to review if it's property set
Image

Questions:

- Is `Gtk.Window.set_wmclass()` necessary? You left a comment that some WM still requires it but which WM needs this for what?
  - https://github.com/gylangsatria/clipboard-history-elementaryos/blob/e41882d8e1c841a8e3379d329b39fe5721fa0c6c/src/window.vala#L36-L38

gylangsatria:
> I'm sorry for my late review. Leaving additional five changes requests, a recommendation, and a question.
> 
> Changes requests:
> 
>     * Your application name `Clipboard History` is too generic. Consider renaming to a unique name, e.g. `ClipStory` (a simple coined word of clipboard and history).
>       
>       * https://docs.elementary.io/develop/appcenter/publishing-requirements#naming
> 
>     * You shouldn't refer your application is for elementary OS because it is assumed that all apps submitted to AppCenter are designed for elementary OS
>       
>       * https://docs.elementary.io/develop/appcenter/publishing-requirements#metadata-requirements
> 
>     * You should not use the `-elementaryos` suffix in your repository name `clipboard-history-elementaryos`
>       
>       * https://docs.elementary.io/develop/appcenter/publishing-requirements#naming
> 
>     * I think `com.github.gylangsatria.clipboard-history-autostart.desktop` wouldn't work as expected
>       
>       * The correct way that supports Flatpak is to use Portals
>       * See https://github.com/elementary/mail/pull/882/changes for example of how to do it (but note that `libportal` is already included in the elementary Flatpak runtime and you don't need to add it in your Flatpak manmifest
> 
>     * Please make sure to create a new tag and update not only the `commit` tag but also the `version` tag in the JSON file of this branch when there are changes in your source code
> 
> 
> Recommendations:
> 
>     * Now that it is recommended to use `io.github` prefix instead of `com.github` because the latter is reserved for GitHub itself in Flathub and thus your app may not be accepted if you want to submit it there
>       
>       * https://docs.flathub.org/docs/for-app-authors/requirements#application-id
> 
>     * The application icon in the about dialog is missing in my installation, even if I restart my system. I would recommend you to review if it's property set
> 
> 
> Image
> 
> Questions:
> 
>     * Is `Gtk.Window.set_wmclass()` necessary? You left a comment that some WM still requires it but which WM needs this for what?
>       
>       * https://github.com/gylangsatria/clipboard-history-elementaryos/blob/e41882d8e1c841a8e3379d329b39fe5721fa0c6c/src/window.vala#L36-L38

Thank you for the thorough review! All change requests have been addressed.

First of all, thank you for suggesting the name ClipStory. It's a great name and I really appreciate it.

I've renamed the application to ClipStory, including all desktop files, metainfo, about dialog, and version labels. References to "for elementary OS" have been removed from descriptions and summaries as requested. I've also renamed the repository from clipboard-history-elementaryos to clipstory.

For autostart, I replaced the old .desktop file approach with Xdp.Portal.request_background() using xdg-desktop-portal. The libportal and libportal-gtk3 dependencies have been added to [meson.build](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html). The deprecated set_wmclass() has been removed since it's not needed and causes compiler warnings. I also fixed the About dialog icon by changing set_logo_icon_name("clipboard-history") to set_logo_icon_name("com.github.gylangsatria.clipboard-history").

Regarding the io.github prefix recommendation, I'll keep that in mind if I decide to submit to Flathub in the future. For now, I'll stick with com.github for AppCenter.

I've pushed all changes and created tag v1.5.0. The JSON file in this PR has been updated with the new commit SHA, version, and source URL.

Please take another look when you have time. Thanks!

applications/com.github.gylangsatria.clipboard-history.json
```diff
@@ -0,0 +1,5 @@
+{
+  "source": "https://github.com/gylangsatria/clipstory",
+  "commit": "024ebe36c9909bcafbd896352c84945edb1f063c",
+  "version": "1.5.0"
```
ryonakano:
```suggestion
  "commit": "910fc92785db8e04514e5942d5995ffe3c8b1764",
  "version": "v1.5.0"
```

As far as looking at https://github.com/gylangsatria/clipstory/commit/910fc92785db8e04514e5942d5995ffe3c8b1764 I think this is the correct SHA and tag.

ryonakano (changes requested):
Changes requests:

- In previous reviews I forgot to say that screenshots of your app should only include the app window itself, not the whole desktop
- Would you also rename your application id from `com.github.gylangsatria.clipboard-history` to `com.github.gylangsatria.clipstory`?
- `Clipboard History` remains in some places. Are they expected?
```
ryonakano@fedora-host:~/work/tmp/clipstory ((v1.5.0))$ shopt -s extglob
ryonakano@fedora-host:~/work/tmp/clipstory ((v1.5.0))$ git grep -n "Clipboard History" -- !(CHANGELOG.md)
README.md:81:  Clipboard History - Dev Tools
data/com.github.gylangsatria.clipboard-history.metainfo.xml:50:      <caption>Clipboard History in dark mode</caption>
data/com.github.gylangsatria.clipboard-history.metainfo.xml:54:      <caption>Clipboard History in light mode</caption>
deb-package/etc/xdg/autostart/com.github.gylangsatria.clipboard-history-autostart.desktop:2:Name=Clipboard History
Binary file deb-package/usr/bin/com.github.gylangsatria.clipboard-history matches
deb-package/usr/share/applications/com.github.gylangsatria.clipboard-history.desktop:2:Name=Clipboard History
dev-setup.sh:222:    echo "  Clipboard History - Dev Tools"
ryonakano@fedora-host:~/work/tmp/clipstory ((v1.5.0))$
```
- I think `data/com.github.gylangsatria.clipboard-history-autostart.desktop` can be deleted in favor of Portal
- Please make sure values of `commit` and `version` is correct (I commented below)

Future suggestions:

- In future releases, please add a new `<release>` tag with new description instead of modifying the old releases in the metainfo file
```diff
   <releases>
-    <release version="1.4.5" date="2026-07-14">
+    <release version="1.5.0" date="2026-07-20">
       <description>
         <p>AppCenter readiness: new Application ID, Flatpak manifest, AppStream metainfo, and file restructuring.</p>
       </description>
```

applications/com.github.gylangsatria.clipboard-history.json
gylangsatria: Thanks for the review. I've addressed all of the requested changes:

## 1. Screenshots
The screenshots have been updated to show only the application window (1126×848) instead of the entire desktop.

## 2. Application ID
The application ID has been renamed from `com.github.gylangsatria.clipboard-history` to `com.github.gylangsatria.clipstory` throughout the project, including:

- `.desktop`
- `.metainfo.xml`
- `.yml`
- Icons and assets
- `meson.build`
- `src/main.vala`
- `src/window.vala`
- `deb-package/`

## 3. "Clipboard History" References
All remaining references to "Clipboard History" have been replaced with "ClipStory", including:

- The menu title in `README.md`
- Screenshot captions in the AppStream metadata

## 4. Autostart Desktop File
The autostart desktop files have been removed:

- `data/com.github.gylangsatria.clipboard-history-autostart.desktop`
- `deb-package/etc/xdg/autostart/com.github.gylangsatria.clipboard-history-autostart.desktop`

`meson.build` has also been updated so these files are no longer installed, since the application now uses the Background Portal instead.

## 5. Commit and Version
These changes will be included in the upcoming **v1.5.1** release. I'll update the AppCenter submission with the new commit hash after pushing the changes.

## 6. Future Releases
Understood. Future AppStream updates will add new `<release>` entries instead of modifying existing ones.

Thanks again for taking the time to review my submission and for all the helpful feedback. I really appreciate it, and it has helped improve the project. If you notice anything else that should be changed, I'd be happy to update it.

ryonakano (changes requested):
Almost there, otherwise LGTM.

One changes request:

- Please make sure to add a `<release>` tag when you release a new version of the app
  - `<release>` tag of v1.5.1 is missing at the moment
  - https://github.com/gylangsatria/clipstory/blob/7c00e5e2f469c3d8e1039b28c5ed6dd98ae576eb/data/com.github.gylangsatria.clipstory.metainfo.xml#L59-L60

One suggestion:

- There is an small error in the metainfo file that the custom `Purism::form_factor` key is duplicated  
  ```
  user@elementary-8-daily:~/clipstory$ LANG=C.UTF-8 appstreamcli validate builddir/export/share/metainfo/com.github.gylangsatria.clipstory.metainfo.xml 
  E: com.github.gylangsatria.clipstory:105: custom-key-duplicated Purism::form_factor
  
  ✘ Validation failed: errors: 1
  user@elementary-8-daily:~/clipstory$
  ```
  - https://github.com/gylangsatria/clipstory/blob/7c00e5e2f469c3d8e1039b28c5ed6dd98ae576eb/data/com.github.gylangsatria.clipstory.metainfo.xml#L104-L105
  - Anyways you should use the official key name `<recommends>` and `<requires>` instead of the custom one
    - e.g. https://gitlab.gnome.org/GNOME/gnome-weather/-/merge_requests/159

gylangsatria:
> Almost there, otherwise LGTM.
> 
> One changes request:
> 
>     * Please make sure to add a `<release>` tag when you release a new version of the app
>       
>       * `<release>` tag of v1.5.1 is missing at the moment
>       * https://github.com/gylangsatria/clipstory/blob/7c00e5e2f469c3d8e1039b28c5ed6dd98ae576eb/data/com.github.gylangsatria.clipstory.metainfo.xml#L59-L60
> 
> 
> One suggestion:
> 
>     * There is an small error in the metainfo file that the custom `Purism::form_factor` key is duplicated
>       ```
>       user@elementary-8-daily:~/clipstory$ LANG=C.UTF-8 appstreamcli validate builddir/export/share/metainfo/com.github.gylangsatria.clipstory.metainfo.xml 
>       E: com.github.gylangsatria.clipstory:105: custom-key-duplicated Purism::form_factor
>       
>       ✘ Validation failed: errors: 1
>       user@elementary-8-daily:~/clipstory$
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       
>       * https://github.com/gylangsatria/clipstory/blob/7c00e5e2f469c3d8e1039b28c5ed6dd98ae576eb/data/com.github.gylangsatria.clipstory.metainfo.xml#L104-L105
>       * Anyways you should use the official key name `<recommends>` and `<requires>` instead of the custom one
>         
>         * e.g. https://gitlab.gnome.org/GNOME/gnome-weather/-/merge_requests/159

Thank you for the detailed review, both issues have been addressed in v1.5.1.

## 1. Release Tag v1.5.1
Added `<release version="1.5.1" date="2026-07-22">` to `com.github.gylangsatria.clipstory.metainfo.xml`

## 2. Purism::form_factor Duplicate
Removed the duplicate custom key, replaced with standard `<recommends>`:

```xml
<recommends>
  <id>org.freedesktop.formfactor.workstation</id>
  <id>org.freedesktop.formfactor.mobile</id>
</recommends>
```

ryonakano:
@gylangsatria Would you tell me the website(s) you referred when you made the following change?

I couldn't find any websites that says `org.freedesktop.formfactor.workstation` and `org.freedesktop.formfactor.mobile` as a valid values, including in [the freedesktop.org documentation](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html).

```xml
<recommends>
  <id>org.freedesktop.formfactor.workstation</id>
  <id>org.freedesktop.formfactor.mobile</id>
</recommends>
```

applications/com.github.gylangsatria.clipstory.json
```diff
@@ -0,0 +1,5 @@
+{
+  "source": "https://github.com/gylangsatria/clipstory",
+  "commit": "7c00e5e2f469c3d8e1039b28c5ed6dd98ae576eb",
+  "version": "v1.5.1"
```
ryonakano: `v1.5.1` does not include the latest change.

```
ryonakano@fedora-host:~/work/tmp/clipstory (main =)$ git log -2
commit 01f3d5e6c8a52609d917416d16c48bd6df83c0aa (HEAD -> main, origin/main, origin/HEAD)
Author: gylangsatria <gylangsatriayudha@gmail.com>
Date:   Wed Jul 22 20:28:50 2026 +0700

    metainfo: add release v1.5.1, replace duplicate Purism::form_factor with <recommends>

commit 7c00e5e2f469c3d8e1039b28c5ed6dd98ae576eb (tag: v1.5.1)
Author: gylangsatria <gylangsatriayudha@gmail.com>
Date:   Wed Jul 22 09:36:32 2026 +0700

    fix: apply AppCenter review feedback

    - Rename Application ID from com.github.gylangsatria.clipboard-history to com.github.gylangsatria.clipstory
    - Remove autostart desktop files (replaced by Portal)
    - Fix remaining 'Clipboard History' references to 'ClipStory'
    - Update screenshots to show app window only
    - Update captions in metainfo.xml
    - Update all file paths and references
ryonakano@fedora-host:~/work/tmp/clipstory (main =)$
```

ryonakano: @gylangsatria If you'd like to release this as v1.5.1, please delete the current tag and recreate it for `01f3d5e6c8a52609d917416d16c48bd6df83c0aa`.

gylangsatria: Sorry for the mistake. I've deleted the old tag and recreated it at `01f3d5e`. The PR should be correct now. Thanks for the review!

ryonakano:
@gylangsatria Would you answer my question at https://github.com/elementary/appcenter-reviews/pull/687#issuecomment-5047185294? Thanks!

gylangsatria:
> @gylangsatria Would you tell me the website(s) you referred when you made the following change?
> 
> I couldn't find any websites that says `org.freedesktop.formfactor.workstation` and `org.freedesktop.formfactor.mobile` as a valid values, including in [the freedesktop.org documentation](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html).
> 
> ```
> <recommends>
>   <id>org.freedesktop.formfactor.workstation</id>
>   <id>org.freedesktop.formfactor.mobile</id>
> </recommends>
> ```

I'm sorry for the late reply!

Thanks for the review and for the thoughtful question. 🙏

To be completely honest, **I couldn't find any official documentation** (including the freedesktop.org one you linked) that explicitly lists `org.freedesktop.formfactor.workstation` and `org.freedesktop.formfactor.mobile` as valid values.

So where did I get them from?

I took the approach from **existing implementations in other GNOME projects**, specifically:

1. **GNOME Weather Merge Request (!159)** — they replaced the deprecated `Purism::form_factor` with the exact same `<recommends>` block:
   - https://gitlab.gnome.org/GNOME/gnome-weather/-/merge_requests/159

2. **GNOME Discourse Discussion (2023)** — community discussion about migrating away from Purism custom keys:
   - https://discourse.gnome.org/t/appdata-custom-purism-tags/17544

The logic is:
- Purism originally used values "workstation" and "mobile" in their custom key
- When migrating to the standard `<recommends>` tag, the community converted those values to proper reverse-DNS IDs (`org.freedesktop.formfactor.workstation` and `org.freedesktop.formfactor.mobile`)
- This follows the same pattern as other well-known IDs like `org.freedesktop.Platform` or `org.freedesktop.Sdk`

I understand the concern about official documentation. This approach is based on **community consensus and existing practice**, not on explicit documentation. That's why I'm referencing GNOME's actual codebase as the source.

If you think we should use a different approach or if you have a better reference, I'm open to suggestions! 

The change does pass validation without errors:

```bash
$ appstreamcli validate data/com.github.gylangsatria.clipstory.metainfo.xml
Validation was successful: 0 errors, 0 warnings
```

ryonakano:
Thank you for your detailed reply.

> When migrating to the standard `<recommends>` tag, the community converted those values to proper reverse-DNS IDs (`org.freedesktop.formfactor.workstation` and `org.freedesktop.formfactor.mobile`)

This is the thing I'm concerning about that the code you fixed may not work as expected.

- The author of https://gitlab.gnome.org/GNOME/gnome-weather/-/merge_requests/159 did not fix in the same way; they used `<control>` and `<display_length>` tags instead of `<id>`
- [The freedesktop.org documentation](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-id) says `<id>` is for representing a relation to another software component
  - On the other hand, `Purism::form_factor` is for flagging which type of devices the application supports, as per in [the blog post by Purism](https://puri.sm/posts/specify-form-factors-in-your-librem-5-apps/)

For these reasons, I think the replacement of deprecated `Purism::form_factor` is `<control>` (and `<display_length>`) tags instead of `<id>`. What do you think?

---

Actually, this is completely out of scope of AppCenter review and not a release blocking problem since the appstream warning is now gone as you mentioned (just out of concern that this may not work as expected) so I'll approve your submission as it is now if you want.

gylangsatria:
> Thank you for your detailed reply.
> 
> > When migrating to the standard `<recommends>` tag, the community converted those values to proper reverse-DNS IDs (`org.freedesktop.formfactor.workstation` and `org.freedesktop.formfactor.mobile`)
> 
> This is the thing I'm concerning about that the code you fixed may not work as expected.
> 
> * The author of https://gitlab.gnome.org/GNOME/gnome-weather/-/merge_requests/159 did not fix in the same way; they used `<control>` and `<display_length>` tags instead of `<id>`
> * [The freedesktop.org documentation](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-id) says `<id>` is for representing a relation to another software component
>   
>   * On the other hand, `Purism::form_factor` is for flagging which type of devices the application supports, as per in [the blog post by Purism](https://puri.sm/posts/specify-form-factors-in-your-librem-5-apps/)
> 
> For these reasons, I think the replacement of deprecated `Purism::form_factor` is `<control>` (and `<display_length>`) tags instead of `<id>`. What do you think?
> 
> Actually, this is completely out of scope of AppCenter review and not a release blocking problem since the appstream warning is now gone as you mentioned (just out of concern that this may not work as expected) so I'll approve your submission as it is now if you want.

Thank you for the clarification and for the detailed explanation! You raise a very valid point, and I appreciate you digging into this.

After reading your comment and re-examining the references:

1. **GNOME Weather MR (!159)** — indeed uses `<control>` and `<display_length>`, not `<id>`. I initially referenced it as evidence of migrating away from `Purism::form_factor`, but I see now that the implementation approach is different.

2. **Your point about `<id>`** — you're absolutely right. The freedesktop.org documentation states that `<id>` is for representing a **relation to another software component**, not for declaring form factor support. And you're correct that the more appropriate replacement would be:

```xml
<control>
  <id>workstation</id>
  <display_length>...</display_length>
</control>
```
or something similar that properly maps to the form factor concept.


Honestly, I followed what I saw in other projects that made similar changes, but I may have misinterpreted the approach. The key thing is:

- The validation error (`custom-key-duplicated`) is now gone 
- The app should work regardless of this metadata tag 
- This is a metadata/declaration concern, not runtime behavior 

Since you mentioned this is not a release-blocking issue and the validation passes, I'd be happy to:

**Option 1:** Keep it as-is for now (since it's working and passes validation), and I can open a follow-up issue to research the proper `<control>` approach later.

**Option 2:** Investigate the `<control>` approach further and update it before merging.

Since you've kindly offered to approve as-is, I think **Option 1** makes sense to avoid delaying the release. I'll make sure to revisit this properly in a future PR.

Let me know if you prefer Option 2 instead — I'm happy to do either! 

Thank you again for the thoughtful review!

ryonakano:
> Since you've kindly offered to approve as-is, I think **Option 1** makes sense to avoid delaying the release. I'll make sure to revisit this properly in a future PR.

Yes, I'm fine with this option. I'll merge this PR if you're fine too.

> And you're correct that the more appropriate replacement would be:

I guess something like this would work:

```xml
  <supports>
    <control>pointing</control>
    <control>keyboard</control>
    <control>touch</control>
  </supports>

  <requires>
    <!-- taken from default_width of MainWindow -->
    <display_length side="shortest" compare="ge">420</display_length>
    <!-- taken from default_height of MainWindow -->
    <display_length side="longest" compare="ge">500</display_length>
  </requires>
```

FYI I'm doing the same thing in my app:

https://github.com/ryonakano/reco/blob/e51957694d54cd9cdaa77cd643f048bc097a4965/data/reco.metainfo.xml.in.in#L48-L59

and it works as expected like this in GNOME Software on a desktop PC:

image

and on a mobile phone (postmarketOS):

Screenshot from 2026-07-24 07:12:27

ryonakano (approved):

